PR #481, opened 2015-01-19
Export for an FOI request

jamiecobbett:
This is based on @issyl0's work.

It (currently) has to be run within the `script/export_for_foi` directory.

I've tested this in development with the most recent database dump:
```
➜  export_for_foi git:(foi_export) ✗ time ./export.sh
Started site export
Finished site export
Started hosts export
Finished hosts export
Started mappings export
Finished mappings export
./export.sh  1.30s user 1.92s system 18% cpu 17.593 total

➜  export_for_foi git:(foi_export) ✗ wc -l *.csv
     2965 hosts-2015-01-19T13:29:17Z.csv
  1912341 mappings-2015-01-19T13:29:17Z.csv
      689 sites-2015-01-19T13:29:17Z.csv
  1915995 total
```


script/export_for_foi/export_mappings.sql
```diff
@@ -0,0 +1,6 @@
+COPY (
+  SELECT sites.abbr, path, new_url, suggested_url, archive_url,
+    CASE type WHEN 'unresolved' THEN 410 WHEN 'archive' THEN 410 WHEN 'redirect' THEN 301 END AS http_status
+    FROM mappings
+    INNER JOIN sites ON mappings.site_id = sites.id
```
jamiecobbett: We should exclude mappings for sites with a global status with a `WHERE` clause.

script/export_for_foi/export_sites.sql
```diff
@@ -0,0 +1,7 @@
+COPY (
+  SELECT abbr, launch_date,
```
jamiecobbett: We should include the significant querystring parameters

jamiecobbett: Rebased to include.

script/export_for_foi/export.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+timestamp=$(date -u +"%Y-%m-%dT%H:%M:%SZ")
+
+echo 'Started site export'
+psql -d transition_development -f ./export_sites.sql > ./sites-$timestamp.csv
```
jennyd: Postgres authentication won't work like this in production - in dev this is relying on peer auth as the `vagrant` role. Making this into a rake task is probably the easiest way of creating all three CSVs at once. Or I can do it manually...

jamiecobbett: I think as long as the data in development is recent (a few days old) I think it's totally reasonable to run the export in development.

script/export_for_foi/export_mappings.sql
```diff
@@ -0,0 +1,7 @@
+COPY (
+  SELECT sites.abbr, path, new_url, suggested_url, archive_url,
```
jennyd: I think we should `archive_url AS custom_archive_url` here, and then explain in the accompanying text how default archive URLs are constructed, with a [link to Bouncer doing that](https://github.com/alphagov/bouncer/blob/568ece01cefe536d5439fa39fe6df4a403539e6e/lib/bouncer/request_context.rb#L51). That means we can be more explicit about which hostname is used and that the request path and querystring are used, and avoids inflating the data inaccurately.

script/export_for_foi/export_sites.sql
jennyd: We've forgotten the site's homepage - that needs to go in.

script/export_for_foi/export_mappings.sql
jamiecobbett: Agreed. I've added friendlier names to the fields.

script/export_for_foi/export_sites.sql
```diff
@@ -0,0 +1,10 @@
+COPY (
+  SELECT abbr,
+    launch_date AS "Launch Date",
```
jennyd: We've forgotten the homepage - that needs to be included.

script/export_for_foi/export_hosts.sql
```diff
@@ -0,0 +1,6 @@
+COPY (
+  SELECT sites.abbr, hostname
```
jennyd: Let's do `sites.abbr AS "Site abbreviation"` in all three cases, to be more explicit.

script/export_for_foi/export_sites.sql
jamiecobbett: Done.

script/export_for_foi/export_hosts.sql
jamiecobbett: Done

script/export_for_foi/export_sites.sql
```diff
@@ -0,0 +1,12 @@
+COPY (
+  SELECT
+    abbr AS "Abbreviation",
+    launch_date AS "Launch Date",
+    homepage AS "New Homepage",
+    query_params AS "Significant querystring parameters (colon separated)",
+    CASE global_type WHEN 'unresolved' THEN 410 WHEN 'archive' THEN 410 WHEN 'redirect' THEN 301 END AS global_http_status,
+    global_new_url AS "New URL for global 301",
+    CASE global_redirect_append_path WHEN 'f' THEN 'false' WHEN 't' THEN 'true' END AS "Append requested path to new path? (for global 301)"
```
jennyd: I think `"Append requested path to new URL? (for global 301)"` would be more accurate here.

script/export_for_foi/export_sites.sql
```diff
@@ -0,0 +1,12 @@
+COPY (
+  SELECT
+    abbr AS "Abbreviation",
```
jennyd: We also need the site's `tna_timestamp`.

script/export_for_foi/export_sites.sql
jamiecobbett: Done

script/export_for_foi/export_sites.sql
jamiecobbett: Done.

script/export_for_foi/export_hosts.sql
```diff
@@ -0,0 +1,8 @@
+COPY (
+  SELECT
+    sites.abbr AS "Abbreviation",
+    hostname
+    FROM hosts
+    INNER JOIN sites ON hosts.site_id = sites.id
+    ORDER BY sites.abbr, hostname
```
jennyd: We should exclude the aka domains - they're an implementation detail.

jamiecobbett: I thought about this but settled on including them as they are (generally) working domains and some partial transitions use them to serve real users. On second thoughts though, you are right.

jamiecobbett: Done.

jennyd:
This is now a Rake task, which fixes the postgres authentication issue and also means it can be run from the root of the project.

I'd like to deploy this to Preview tomorrow morning to get the data out; I can deploy the branch if this hasn't been merged.

PR #483, opened 2015-02-10
Update the whitelist request email address to reflect the new teams

issyl0:
- For the moment, set this to be govuk-core-team@ instead of
  2nd-line-support@, or transition-dev@ as it was before, to avoid
  these requests only being dealt with by people subscribed to the now
  disbanded Transition Tool development team list - Jamie, Jenny, me,
  etc.

Hopefully soon we'll get around to writing some documentation in the opsmanual for this process and why it is like it is, so that 2nd line can take it on...

(cc @jamiecobbett, @jennyd)

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -63,6 +63,6 @@ module Transition
     config.assets.precompile += %w(html5.js)
     config.assets.precompile += %w(respond.min.js)
 
-    config.support_email = 'transition-dev@digital.cabinet-office.gov.uk'
+    config.support_email = 'govuk-core-team@digital.cabinet-office.gov.uk'
   end
 end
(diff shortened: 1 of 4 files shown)


jamiecobbett:
Perhaps it should go to Zendesk instead?

issyl0:
Following a conversation with Jenny around the Zendesk comment and whether or not it's for the Core Team, I spoke to Steph as the second-line delivery manager. She suggested that second-line would be a good place for these requests to be sent _if_ we write some documentation on why they exist and how we deal with them. She's going to talk to Albert about the possibility of Zendesk, as Jenny and I thought that people requesting short URLs might not always have Zendesk access.

Going to close this for now until we have a definite decision, and work on some documentation explaining why and how to approve or deny whitelist requests. Any ideas as to where that documentation can go? If it ends up being first-line support people who deal with it through Zendesk, without touching second-line, the opsmanual isn't the best place. And it's not a sensitive process...

PR #484, opened 2015-02-24
Strip trailing whitespace from hosts added to the whitelist

issyl0:
- Upon accidentally adding a host with trailing whitespace this morning, thanks to a failed copy/paste, I realised that the whitelist let it through without stripping the whitespace or erroring. Make it strip the whitespace from the hostname, so that people and Bouncer don't get confused or match on the wrong thing.


issyl0:
The tests pass here now, thanks to Jenny yesterday. :sunny:

PR #485, opened 2015-03-09
Add *.nhs.uk as an automatically allowed ending for a hostname

issyl0:
- Like we do for *.gov.uk and *.mod.uk, *.nhs.uk should not need
  whitelisting. We were getting many requests for *.nhs.uk domains to
  be whitelisted, from PHE's transition, and it's easier to deal with
  them globally than have them ask us to add yet another to the
  whitelist when they find one.

### DIFF
--- a/app/models/whitelisted_host.rb
+++ b/app/models/whitelisted_host.rb
@@ -16,8 +16,8 @@ class WhitelistedHost < ActiveRecord::Base
 
   def hostname_is_not_automatically_allowed_anyway
     return if hostname.nil?
-    if hostname.end_with?('.gov.uk') || hostname.end_with?('.mod.uk')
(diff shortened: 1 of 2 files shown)


jamiecobbett:
You'll need a corresponding change in Bouncer to actually allow these redirects.

I think there should a follow-up task (perhaps a migration) to remove the `.nhs.uk` domains.

From a security point of view, the question we need to ask ourselves is how likely is a .nhs.uk domain likely to be compromised?

adammaddison:
I know nhs.uk domains are strictly controlled by HSCIC
<http://www.hscic.gov.uk>.  Specific queries could be sent to
dnsteam@hscic.gov.uk.

No idea if that's relevant...

PR #486, opened 2015-03-13
Allow adding mappings to *.nhs.uk hosts

issyl0:
- This is needed otherwise the validations of whitelisted hosts disallow mappings to be added for, for example, `somethingnotexplicitlywhitelistedyet.nhs.uk` A good spot by Jenny pre-deploy of https://github.com/alphagov/transition/pull/485. :sunny:

### DIFF
--- a/app/validators/host_in_whitelist_validator.rb
+++ b/app/validators/host_in_whitelist_validator.rb
@@ -10,7 +10,7 @@ class HostInWhitelistValidator < ActiveModel::EachValidator
   def in_whitelist?(url)
     uri = Addressable::URI.parse(url)
     return false if uri.host.nil?
-    uri.host.end_with?('.gov.uk') || uri.host.end_with?('.mod.uk') || WhitelistedHost.exists?(hostname: uri.host)
+    uri.host.end_with?('.gov.uk') || uri.host.end_with?('.mod.uk') || uri.host.end_with?('.nhs.uk') || WhitelistedHost.exists?(hostname: uri.host)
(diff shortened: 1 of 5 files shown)


jennyd:
I can now edit mappings to redirect to `*.nhs.uk` domains, which is good :sunglasses: 

I think we should add `.nhs.uk` to the list of TLDs in [`Transition::PathOrURL`](https://github.com/alphagov/transition/blob/master/lib/transition/path_or_url.rb#L6-L18). I don't think it makes any difference to the whitelist because that list only applies to old URLs and not new URLs, but not having it in the list may break some things around canonicalization, because `Transition::PathOrURL.starts_with_a_domain?` is used in [`Site.canonical_path`](https://github.com/alphagov/transition/blob/3d5821ad044ac82fc9a645dc66fc9d02fdf44b4c/app/models/site.rb#L67).

We test that `*.mod.uk` domains are valid new URLs in the [`Mapping` model specs](https://github.com/alphagov/transition/blob/master/spec/models/mapping_spec.rb#L148-L152), but I don't think it's necessary to add a test there for `*.nhs.uk`.

jennyd:
Sorry, when I said that I thought we didn't need to test `*.nhs.uk` in the `Mapping` model specs I was assuming that we were testing the validator elsewhere, but we aren't doing that. It would actually be a good idea to add a test case for it.

issyl0:
@jennyd Thanks for the feedback. Added it to `Transition::PathOrURL` and added a test for it.

PR #487, opened 2015-03-20
Remove the specific *.nhs.uk hosts from the whitelist list

issyl0:
- Anything ending in .nhs.uk is now automatically whitelisted. See
  https://github.com/alphagov/transition/pull/485 and
  https://github.com/alphagov/transition/pull/486. This is the tidying
  up migration to remove the hosts we'd manually added to the
  whitelist before we made the global change.

### DIFF
--- a/db/migrate/20150320164433_remove_individual_nhsuk_domains_from_whitelist.rb
+++ b/db/migrate/20150320164433_remove_individual_nhsuk_domains_from_whitelist.rb
@@ -1,0 +1,12 @@
+class RemoveIndividualNhsukDomainsFromWhitelist < ActiveRecord::Migration
+  def up
+    WhitelistedHost.where("hostname LIKE '%.nhs.uk'").each do |whitelisted_host|
+      whitelisted_host.destroy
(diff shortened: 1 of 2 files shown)


db/migrate/20150320164433_remove_individual_nhsuk_domains_from_whitelist.rb
```diff
@@ -0,0 +1,12 @@
+class RemoveIndividualNhsukDomainsFromWhitelist < ActiveRecord::Migration
+  def up
+    WhitelistedHost.where("hostname LIKE '%.nhs.uk%'").each do |whitelisted_host|
```
jennyd: Strictly we shouldn't have the second `%` here, because we only automatically whitelist domains which end with `.nhs.uk`, rather than ones which contain it.

issyl0: Good point, even though we have no domains that _contain_ it per se. Amended.

PR #488, opened 2015-04-22
Update the whitelist and support request email address to cite Zendesk

issyl0:
- The team behind the transition-dev@ mailing list was dissolved a few
  months ago in the GOV.UK restructure. The email address is still
  active and they still receive sporadic whitelist requests for
  domains. These requests are now fielded via second-line support, and
  so should go to Zendesk. Given that people using the Transition Tool
  are inside government departments or agencies, they should know what
  Zendesk is and how to send queries via it, without us having to link
  explicitly to it.

### DIFF
--- a/app/models/bulk_add_batch.rb
+++ b/app/models/bulk_add_batch.rb
@@ -16,8 +16,7 @@ class BulkAddBatch < MappingsBatch
     redirect.validates :new_url, length: { maximum: 2048 }
     redirect.validates :new_url, non_blank_url: { message: I18n.t('mappings.bulk.new_url_invalid') }
     redirect.validates :new_url, host_in_whitelist: {
-      message: I18n.t('mappings.bulk.new_url_must_be_on_whitelist',
-      email: Rails.configuration.support_email)
+      message: I18n.t('mappings.bulk.new_url_must_be_on_whitelist')
     }
     redirect.validates :new_url, not_a_national_archives_url: {
       message: I18n.t('mappings.bulk.new_url_must_not_be_on_tna')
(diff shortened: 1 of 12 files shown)


jamiecobbett:
I think if we can provide a link to Zendesk, we should provide a link.

rjc123:
+1 to links instead of text.

adammaddison:
Just wanted to add a comment for old times' sake...

issyl0:
@rjc123 Hello, stranger!

rgarner:
Build's failing. But +1 to links.

issyl0:
OK. Following conversations, we should be linking to the Support app because that's where internals users submit requests that then get funnelled into our Zendesk. I've amended that commit to take account of this and add a link to the correct section within the support app - "general requests". Now I just need to work out how to make YAML locale files do HTML - http://guides.rubyonrails.org/i18n.html#using-safe-html-translations isn't working.

issyl0:
@rgarner HELLO! :sunny:

jennyd:
:wave: to all you lovely people :rainbow: :link:

jennyd:
I wonder when @fofr, @davidmann and @fatbusinessman are going to join in...

fatbusinessman:
> I wonder when @fofr, @davidmann and @fatbusinessman are going to join in...

You called?

issyl0:
The build on this is still failing. I'll fix it tomorrow morning once I've worked out what's going on with links in YAML files, as per an earlier comment... :confounded:

issyl0:
The tests now pass - I found the right combination of `.html_safe` and added it to the right file. On Jenny's advice I also added a Cucumber feature to check that the flash message actually has a link in it...

screen shot 2015-04-23 at 16 15 26

fofr:
Yay :fireworks:

adammaddison:
\o/

issyl0:
Goodbye, remnant of the old team... :cloud:

jamiecobbett:
200_s

rjc123:
images

PR #489, opened 2015-04-23
Groundwork for organisation content ids

jamiecobbett:
Set the app up to start recording organisation content IDs against users and organisations.

Once these fields are populated, we can start using them, instead of volatile slugs to connect things together.

It probably makes sense to also switch transition-config to content_ids, but that may make the site config files a bit opaque...


db/migrate/20150421161752_add_disabled_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddDisabledToUser < ActiveRecord::Migration
+  def change
+    add_column :users, :disabled, :boolean
```
jennyd: The [gds-sso README says](https://github.com/alphagov/gds-sso/blob/master/README.md#integration-with-a-rails-3-app) that this column should default to `false`.

jamiecobbett: Good spot. I've also added the shared_example from gds-sso to provide some validation of the model, although it didn't catch this particular problem.

jamiecobbett:
@jennyd I think I've finally undone the effects of trying to rebase with migrations already run. We should just keep the database in git too...

jennyd:
nooooo... now you're missing the final migration timestamp in the schema in the last commit, so the build thinks that there are pending migrations :(

jamiecobbett:
@jennyd NOW I think I've unpicked it.

I feel like
i hate manure _fda717850ed699a6083f6175365a4d06

db/schema.rb
```diff
@@ -214,11 +215,13 @@
...
-    t.boolean  "remotely_signed_out", default: false
-    t.datetime "created_at",                          null: false
-    t.datetime "updated_at",                          null: false
+    t.boolean  "remotely_signed_out",     default: false
+    t.datetime "created_at",                              null: false
+    t.datetime "updated_at",                              null: false
     t.string   "organisation_slug"
-    t.boolean  "is_robot",            default: false
+    t.boolean  "is_robot",                default: false
+    t.string   "organisation_content_id"
```
jennyd: One last thing: the `organisation_content_id`and `disabled` lines should be the other way around - I get a schema diff in dev at the moment after running migrations.

rgarner:
You've got to have a special sort of a brane to get migrations. Preferably one under 40 that is not sleep-deprived in the same way as one who is in some sort of family unit with some sort of recent arrival.

db/schema.rb
jamiecobbett: :goberserk:

jennyd:
After discussing what to do about transition-config, @jamiecobbett and I think we should:

- switch to using `content_id` as the organisation identifiers in site YAML files
- comment each one with the name of the org so that there's a human-readable indication of the ownership of the site
- add a script in transition-config to go through all the site YAML files and update those comments to keep them up-to-date with the current `content_id`s to make the process less error-prone
- validate `content_id`s and the comments as part of the bulid, as we currently do for slugs
- we would either expect users to find the correct `content_id` for the organisation to put into the site YAML file (they are visible in Whitehall's organisations API) or accept the slug and then convert it straight away to content_id automatically (perhaps via the new_site rake task)

All of that can only be done once Transition is fully using `content_id` to identify orgs in the organisations import from Whitehall, so that duplicate orgs are not created when slugs change.

The advantage to all this is that there would be no manual work required in Transition when organisation slugs change; at the moment there are several steps which must be done in the correct order. The cost would be more complexity in transition-config. Having the ability to create and edit site config in the app would avoid that complexity, but would be a complex feature to build because of the amount of domain knowledge we expect site configurers to have (and we would need to explain in the app).

jamiecobbett:
I've started the work in transition-config here: https://github.com/alphagov/transition-config/pull/1090

PR #490, opened 2015-04-28
Remove fake businesslink and directgov orgs

jamiecobbett:
The Directgov and Businesslink organisations were a fudge to provide a separate
list of sites.

Now that all of the sites associated with these fake organisations have been
moved to other organisations, we can remove this complexity.

By doing this, we make the code simpler and we make it easier to switch to
using content IDs as the primary key for organisations.


spec/lib/transition/import/site_yaml_file_spec.rb
```diff
@@ -13,6 +13,7 @@
 
     describe '#import!' do
       let(:ago)  { build :organisation, whitehall_slug: 'attorney-generals-office' }
+      let!(:gds)  { build :organisation, whitehall_slug: 'government-digital-service' }
```
jennyd: I can't see this being used anywhere in this PR...

jamiecobbett: Correct - I added it before using a different organisation instead. Rebased.

PR #491, opened 2015-04-29
Handle organisation slug changes automatically

jamiecobbett:


lib/transition/import/site_yaml_file.rb
```diff
@@ -70,6 +70,10 @@ def global_redirect_append_path
       attr_reader :site
       def import_site!
         @site = Site.where(abbr: abbr).first_or_initialize.tap do |site|
+          # transition-config uses slugs to identify organisations because
+          # content_ids are user-unfriendly and add complexity. We think it will
+          # be very infrequent that an organisation with a slug change will be
+          # in transition-config.
```
jennyd: Nice documentation :snail: :no_entry_sign:  :shell:

app/models/organisation.rb
```diff
@@ -25,6 +25,7 @@ class Organisation < ActiveRecord::Base
   validates_presence_of :whitehall_slug
   validates_uniqueness_of :whitehall_slug
   validates_presence_of :title
+  validates_presence_of :content_id
```
jennyd: We should probably also validate uniqueness of `content_id` and have a unique index on it for consistency (I missed that in https://github.com/alphagov/transition/pull/489). [Null values are not considered equal](http://www.postgresql.org/docs/9.3/static/indexes-unique.html) by Postgres so the orgs in the database which no longer exist by slug in Whitehall and therefore won't get their content_ids set on import shouldn't cause problems.

jamiecobbett: I've added a migration which removes the orphaned organisations and adds a unique constraint to the database. I'm always hesitant with adding `validates_uniqueness_of` as it means ActiveRecord does a query every time you save to get all the values for that field.

PR #497, opened 2015-07-21
Allow defining custom archive URLs in a bulk import

mikejustdoit:
A Site has many Mappings and user can bulk import mappings from a CSV.

One of the available types of Mappings is an Archive Mapping. An Archive Mapping has an Old URL and an Archive URL.

A Site has a National Archives timestamp and its Archive Mappings can either construct their Archive URL using that timestamp or they can specify their own custom Archive URL.

Until now we've ignored any custom Archive URLs that were defined in an imported CSV (and instead let the imported Archive Mappings default to constructing their Archive URLs from their Site's timestamp as described above).

These updates change that behaviour, so that we now respect custom Archive URLs in CSVs and persist them along with their Mapping.

### DIFF
--- a/app/models/import_batch.rb
+++ b/app/models/import_batch.rb
@@ -30,6 +30,11 @@ class ImportBatch < MappingsBatch
     validator: HostInWhitelistValidator,
     message: I18n.t('mappings.bulk.new_url_must_be_on_whitelist')
   }
+  validates :archive_urls, each_in_collection: {
+    validator: LengthValidator,
+    maximum: (64.kilobytes - 1),
+    message: I18n.t('mappings.new_url_too_long')
+  }
 
   after_create :create_entries
 
@@ -38,7 +43,13 @@ class ImportBatch < MappingsBatch
     existing_mappings = site.mappings.where(path: canonical_paths)
 
     deduplicated_csv_rows.each do |row|
-      entry = ImportBatchEntry.new(path: row.path, type: row.type, new_url: row.new_url)
+      entry = ImportBatchEntry.new(
+        path: row.path,
+        type: row.type,
+        new_url: row.new_url,
+        archive_url: row.archive_url,
+      )
+
       entry.mappings_batch = self
       entry.mapping = existing_mappings.detect { |mapping| mapping.path == row.path }
       entry.save!
@@ -51,6 +62,10 @@ class ImportBatch < MappingsBatch
 
   def new_urls
     deduplicated_csv_rows.select(&:redirect?).map(&:new_url).uniq
+  end
+
+  def archive_urls
+    deduplicated_csv_rows.map(&:archive_url).compact
   end
 
   def canonical_paths
--- a/app/views/import_batches/preview.html.erb
+++ b/app/views/import_batches/preview.html.erb
@@ -27,7 +27,12 @@
   <h4 class="add-bottom-margin">Running this import will</h4>
   <ul class="list-group">
     <li class="list-group-item">Create <%= pluralize(@preview.redirect_count, 'new redirect') %></li>
-    <li class="list-group-item">Create <%= pluralize(@preview.archive_count, 'new archive') %></li>
+    <li class="list-group-item">
+      Create <%= pluralize(@preview.archive_count, 'new archive') %>
+      <% if @preview.custom_archive_url_count > 0 %>
+        (<%= pluralize(@preview.custom_archive_url_count, 'with custom URL') %>)
+      <% end %>
+    </li>
     <li class="list-group-item">Create <%= pluralize(@preview.unresolved_count, 'new unresolved mapping') %></li>
     <% if @preview.existing_mappings_count > 0 %>
       <li class="list-group-item list-group-item-warning js-overwrite-count">Overwrite <%= pluralize(@preview.existing_mappings_count, 'existing mapping') %></li>
--- a/features/step_definitions/mappings_import_assertion_steps.rb
+++ b/features/step_definitions/mappings_import_assertion_steps.rb
@@ -11,6 +11,11 @@ Then(/^I should see how many of each type of mapping will be created$/) do
   }
 end
 
+Then(/^I should see how many archive mappings will be created and how many have custom URLs$/) do
+  steps %{
+    Then I should see "Create 3 new archives (2 with custom URLs)"
+  }
+end
 
 Then(/^I should see how many mappings will be overwritten$/) do
   steps %{
--- a/lib/transition/import/csv/import_batch_row.rb
+++ b/lib/transition/import/csv/import_batch_row.rb
@@ -53,6 +53,10 @@ module Transition
           archive? && new_value && new_url_is_a_national_archives_url?
         end
 
+        def archive_without_custom_url?
+          archive? && !archive_with_custom_url?
+        end
+
         def redirect?
           type == 'redirect'
         end
@@ -67,6 +71,10 @@ module Transition
           elsif redirect?
             1
           elsif archive? && other.redirect?
+            -1
+          elsif archive_with_custom_url? && other.archive_without_custom_url?
+            1
+          elsif archive_without_custom_url? && other.archive_with_custom_url?
             -1
           elsif archive? && other.archive?
             other.line_number <=> line_number
--- a/lib/transition/import/csv/import_batch_row.rb
+++ b/lib/transition/import/csv/import_batch_row.rb
@@ -41,8 +41,16 @@ module Transition
           new_value if redirect?
         end
 
+        def archive_url
+          new_value if archive_with_custom_url?
+        end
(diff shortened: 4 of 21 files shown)


fofr:
Copy changes look good. It _might_ be worth adding a line to https://github.com/alphagov/transition/blob/master/app/views/import_batches/new.html.erb#L24 to show an example of a custom National Archive.

mikejustdoit:
As discussed in person with @fofr, we're gonna forget adding an example CSV entry for now.

The loooonngg National Archives URL results in the line wrapping and being a lot less readable than the other, existing examples.

And it's an advanced feature that's already explained on the page (https://github.com/alphagov/transition/blob/master/app/views/import_batches/new.html.erb#L17).

spec/support/shared_examples/mappings_batch.rb
```diff
@@ -1,3 +1,34 @@
+shared_examples 'creates redirect mapping' do
+  context 'some context' do
```
rgarner: Any reason we're keeping a couple of contexts with slightly vague names?

(apologies if I shouldn't be looking at this, I should really have unsubscribed ;))

mikejustdoit: Yep, @rgarner, I tried to explain this, possibly not very well, in the commit messages for those changes (last line in the message) https://github.com/alphagov/transition/commit/36461a3b88873e0447101757a42867c01361f589

app/views/import_batches/preview.html.erb
```diff
@@ -27,7 +27,12 @@
   <h4 class="add-bottom-margin">Running this import will</h4>
   <ul class="list-group">
     <li class="list-group-item">Create <%= pluralize(@preview.redirect_count, 'new redirect') %></li>
-    <li class="list-group-item">Create <%= pluralize(@preview.archive_count, 'new archive') %></li>
+    <li class="list-group-item">
+      Create <%= pluralize(@preview.archive_count, 'new archive') %>
+      <% if @preview.custom_archive_url_count > 0 %>
+        (<%= pluralize(@preview.custom_archive_url_count, 'with custom URL') %>)
```
jennyd: We describe this field as 'Alternative National Archives URL' in the [single edit mapping form](https://github.com/alphagov/transition/blob/4f2fe54d6753e4fcff11390c7ba8f91a75a1665e/app/views/mappings/_form.html.erb#L57) - it would be good to keep the wording consistent, although that's a longer description.

app/models/import_batch.rb
```diff
@@ -38,7 +38,13 @@ def create_entries
     existing_mappings = site.mappings.where(path: canonical_paths)
 
     deduplicated_csv_rows.each do |row|
-      entry = ImportBatchEntry.new(path: row.path, type: row.type, new_url: row.new_url)
+      entry = ImportBatchEntry.new(
+        path: row.path,
+        type: row.type,
+        new_url: row.new_url,
+        archive_url: row.archive_url,
```
jennyd: Re your questions in the [commit message](https://github.com/alphagov/transition/commit/59ddcab0bb384e20c390e2acf64db994cc33c864) about validations on the batch, we should probably validate the archive URLs for length, but there's no point in validating for non-blankness or that they're National Archives URLs. They will only be set as archive URLs on the entries if their [hosts matched `NationalArchivesURLValidator:: NATIONAL_ARCHIVES_HOST `](https://github.com/alphagov/transition/blob/4f2fe54d6753e4fcff11390c7ba8f91a75a1665e/lib/transition/import/csv/import_batch_row.rb#L73) which covers both of those cases already.

features/step_definitions/fixture_steps.rb
```diff
@@ -46,6 +46,11 @@
   Host.first.site.delete
 end
 
+Given(/^there is a site called (.*) belonging to an organisation ([^ ]*)$/) do |site_abbr, org_abbr|
+  org = create(:organisation, title: org_abbr, whitehall_slug: org_abbr)
+  create(:site, organisation: org, abbr: site_abbr)
+end
```
jennyd: The existing [`Given(/^a site (.*) exists$/)`](https://github.com/alphagov/transition/blob/a305b6d436ec30ab9146380f08fd9e179cfbcce6/features/step_definitions/fixture_steps.rb#L98-L100) step does pretty much this already, so you should be able to use that instead of introducing a new step. The `site` factory has `association :organisation` so will create an org too, which is enough to make the site dashboard page work.

features/step_definitions/mappings_import_assertion_steps.rb
```diff
@@ -11,6 +11,17 @@
   }
 end
 
+Then(/^I should see how many archive mappings will be created$/) do
+  steps %{
+    Then I should see "Create 3 new archives"
```
jennyd: This step's assertion duplicates part of the following step - is there a way of rewording the description of the latter so that it makes sense to use only that one?

lib/transition/import/csv/import_batch_row.rb
```diff
@@ -60,6 +64,8 @@ def <=>(other)
             1
           elsif archive? && other.redirect?
             -1
+          elsif archive? && other.archive?
+            other.line_number <=> line_number
```
jennyd: I'm not sure whether this behaviour is the right thing to do, because it would be difficult for users to know how to reliably get their custom archive URLs saved.

For example, say a user had noticed that `http://foo.gov.uk/a?x=y` needed a custom archive URL and imported a CSV including these two rows (in this order):

```
http://foo.gov.uk/a?b=c,TNA
http://foo.gov.uk/a?x=y,http://webarchive.nationalarchives.gov.uk/20120816224015/http://foo.gov.uk/a
```

Those rows would both have the same canonical path `/a`, and with the current behaviour the earlier one would take precedence, losing the custom archive URL.

To get around this, the user would have to know that they need to add the custom archive URL to the earliest row in the CSV which would canonicalize to that path in order for it to be saved.

I think it would appear more consistent to users for an archive row with a custom archive URL to take precedence over earlier normal archive rows. If there are multiple rows with custom archive URLs then we should fall back to using the earliest one. This would increase the complexity of the code, but I think it's probably worth it for the more understandable behaviour for users.

jennyd:
Lovely small commits with great explanations at every stage :sparkles: :cherry_blossom: :pizza: :small_blue_diamond:

features/step_definitions/fixture_steps.rb
mikejustdoit: Thanks! This wasn't obvious

app/views/import_batches/preview.html.erb
mikejustdoit: I came across that view while working on this, but I found it hard to understand. Obviously it's wrong to introduce inconsistent wording, but are we sure that the original wording says what it means?

features/step_definitions/mappings_import_assertion_steps.rb
mikejustdoit: Thanks, fixed now.

app/models/import_batch.rb
mikejustdoit: Fixed and added your explanation to the commit message!

lib/transition/import/csv/import_batch_row.rb
mikejustdoit: Thanks, done that now

app/models/import_batch.rb
```diff
@@ -30,6 +30,11 @@ class ImportBatch < MappingsBatch
     validator: HostInWhitelistValidator,
     message: I18n.t('mappings.bulk.new_url_must_be_on_whitelist')
   }
+  validates :archive_urls, each_in_collection: {
+    validator: LengthValidator,
+    maximum: (64.kilobytes - 1),
+    message: I18n.t('mappings.new_url_too_long')
```
jennyd: This message is [specific to new URL](https://github.com/alphagov/transition/blob/4f2fe54d6753e4fcff11390c7ba8f91a75a1665e/config/locales/en.yml#L27), which could be confusing to users. We could make that message more generic, add a separate one for custom archive URLs or make the message take the name of the field to use.

app/views/import_batches/preview.html.erb
jennyd: We've decided to change the single edit mapping form field label to 'Custom National Archives URL'.

app/models/import_batch.rb
mikejustdoit: You're right, it might be confusing to users!

I believe in this case, the user has _just_ finished uploading their CSV of "Old URL,New URL", so there's no distinct field for an 'Archive URL' at this point, and 'New URL' still refers to the column name.

Which is the last in your list of possible solutions - take the name of the field - so I assume this is probably our best option right now?

fofr:
:stars: :star2:

PR #498, opened 2015-08-18
Add a rake task to revert import of some sites

jennyd:
We occasionally configure sites and then later decide that they should not be
transitioned via the Transition tool. In these cases it's good to delete the
imported site so that the data reflects our decisions. An example of such a
site is the content planner application on alphagov.co.uk: this app is [no
longer in Production](https://github.com/alphagov/transition-config/pull/1128) before the move to publishing.service.gov.uk so we don't
need to redirect the old domain, but it has already been imported.

This rake task makes it easier to delete these sites and all associated data
which may have been created by the site import: hosts and links to extra
organisations. The site should be removed from the transition-config repo
before running the task, otherwise the site will be re-imported the next time
the `orgs_sites_hosts` import runs.

The task takes an array of site abbrs, eg:

    bundle exec rake import:revert:sites[gds_alphagov_content-planner,another_site]

If any given site has any mappings or hits rows, it will not be deleted. The
presence of these is strongly indicative of us needing to keep the sites in
the database; if we really do want to delete a site which has mappings or hits
rows then that is a very special case and won't be automated.

### DIFF
--- a/lib/transition/import/revert.rb
+++ b/lib/transition/import/revert.rb
@@ -1,0 +1,54 @@
+require 'transition/import/console_job_wrapper'
+
+module Transition
+  module Import
+    module Revert
+      ##
+      # Reverts the import of sites, if they have no mappings or hits - ie no
+      # associated data beyond that which was created by importing the site
+      # YAML file
+      class Sites
+        include Transition::Import::ConsoleJobWrapper
+
+        def initialize(site_abbrs)
+          @site_abbrs = site_abbrs
+        end
+
+        def revert_all!
+          console_puts "Trying to delete sites: #{@site_abbrs.join(', ')}"
+
+          @site_abbrs.each { |abbr| revert_safely!(abbr) }
+
+          console_puts 'Ensure that the deleted sites have also been deleted from the transition-config repo; otherwise they will be re-imported.'
+        end
+
+      private
+        def revert_safely!(abbr)
+          Site.transaction do
+            site = Site.find_by(abbr: abbr)
+            unless site
+              console_puts "Site #{abbr} doesn't exist; skipping"
+              return
+            end
+
+            mappings_count = site.mappings.count
+            hits_count = site.hits.count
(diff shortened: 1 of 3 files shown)


lib/transition/import/revert.rb
```diff
@@ -0,0 +1,54 @@
...
+            if mappings_count == 0 && hits_count == 0
+              delete_site_and_associations(site)
+              console_puts "Deleted site #{abbr}"
+            else
+              console_puts "Site #{abbr} has #{mappings_count} mappings and #{hits_count} hits; not deleting"
+            end
+          end
+        end
+
+        def delete_site_and_associations(site)
```
jamiecobbett: `delete` and `destroy` have different meanings in ActiveRecord's API - given that this method uses `destroy`, it would be better to name the method the same I think.

lib/transition/import/revert.rb
```diff
@@ -0,0 +1,54 @@
...
+    module Revert
+      ##
+      # Reverts the import of sites, if they have no mappings or hits - ie no
+      # associated data beyond that which was created by importing the site
+      # YAML file
+      class Sites
+        include Transition::Import::ConsoleJobWrapper
+
+        def initialize(site_abbrs)
+          @site_abbrs = site_abbrs.reject(&:empty?)
```
jamiecobbett: The `reject(&:empty?)` seems to me like it is about cleaning up dodgy input - maybe it belongs in the Rake task?

jamiecobbett:
A couple of minor points, otherwise :+1:

lib/transition/import/revert.rb
jennyd: Good point, renamed

lib/transition/import/revert.rb
jennyd: Thanks, I've done that now

jamiecobbett:
Merge when green :green_book:

jamiecobbett:
1849252585_f528d439c3_z

PR #502, opened 2015-09-29
Update alphagov domain name in production

alexmuller:
We've moved domain names in production. The old ones are redirected but we should update them anyway.


issyl0:
Thanks, Alex, though you'll need to update the tests as well. :sunny:

PR #504, opened 2016-01-18
Add rake task to remove site and all data

brenetic:
- ONS domains and data need to be removed from transition as they would like to add a different service

- The rake task takes a site abbreviation as a parameter and removes all associated data from the following tables: 
   - mappings
   - versions
   - daily_hit_totals
   - host_paths
   - sites
   - organisational_relationships
   - organisations_sites

- The hits files in the hits directory are not removed.

https://trello.com/c/WVYEPfxG/252-delete-all-data-for-www-ons-gov-uk-in-transition

### DIFF
--- a/lib/tasks/import/revert_entirely_unsafe.rake
+++ b/lib/tasks/import/revert_entirely_unsafe.rake
@@ -1,0 +1,22 @@
+namespace :import do
+  desc 'Deletes a site and all associated data'
+  task :revert_entirely_unsafe, [:site_abbr] => :environment do |_, args|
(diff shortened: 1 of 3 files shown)


jennyd:
Could you add tests for this new class, as we [have already](https://github.com/alphagov/transition/tree/master/spec/lib/transition/import) for the other classes in the `import` module? It's particularly important to test this because it's destructive and could break other sites in Production if it's not doing the right things.

brenetic:
@jennyd Added a test.

h-lame:
Good stuff.  Merging.

PR #506, opened 2016-01-26
Upgrade rails 4.1.14.1

brenetic:
- Upgraded rails to 4.1.14.1 as a result of https://groups.google.com/forum/#!topic/rubyonrails-security/dthJ5wL69JE
- Bumped Sidekiq to 3.4.2
- Bumped Uglifier to 2.7.2
- Implemented temporary alias to get round the @cache variable not being initialized correctly in EmptyTemplatePathSetDecorator
    - This is taken from the suggestion here: https://github.com/rspec/rspec-rails/issues/1532#issuecomment-174679485
    - This should be removed once RSpec has been patched

https://trello.com/c/Ttk9WA34/40-transition


benlovell:
Was bumping `sidekiq` required for compatibility or otherwise? If otherwise, given the nature of this PR it should probably happen separately.

spec/support/rspec_rails_4.1.14.1.rb
```diff
@@ -0,0 +1,7 @@
+# Fix for rspec-rails and rails 4.1.14.1
+# This can be removed as soon as doing so doesn't cause test errors.
+# See https://github.com/rspec/rspec-rails/issues/1532
+
+RSpec::Rails::ViewRendering::EmptyTemplatePathSetDecorator.class_eval do
```
benlovell: It's worth guarding against the correct version of `rspec-rails` here, and raising an error if there is a mismatch.

brenetic:
Thanks @benlovell fixed the spelling error.

Sidekiq came up as a warning when I ran the security audit against transition hence the upgrade. Same for Uglifier. I can remove them into a separate pull request?

benlovell:
@brenetic I guess it's not a huge deal if they're in this PR, especially since they are separate commits. I'll leave it with you :+1:

spec/support/rspec_rails_4.1.14.1.rb
brenetic: @benlovell Added a check for that. Hopefully the fix will be in the next few days anyway. But also this app is running a fairly old version of rspec-rails.

benlovell:
LGTM! :shipit: when it's green.

spec/support/rspec_rails_4.1.14.1.rb
benlovell: Ah, `3.4.1` was literally _just_ released.

brenetic: Ha! :)

I shall back this out and get 3.4.1

brenetic:
The upgrading of RSpec from 2.14.2 to 3.4.1, and all required dependencies, is non trivial. In the interesting of getting the security patch out it might be better to merge this and then raise another pull request with the necessary updates to RSpec.

benlovell:
@brenetic I never realised it wasn't already on the `3.x` series (I never checked)

That seems smart to me. :ship:

PR #510, opened 2016-05-06
Comment out some migrations so that first time setup runs cleanly

issyl0:
- Setting up Transition on a new laptop, I replicated data and then had
  to run `rake db:migrate`. It failed on the "enable pgcrypto" migration
  because it requires database superuser access to enable extensions. We
  don't need this extension anymore (it was removed in November 2014),
  so comment out migrations that reference it so that `db:migrate` can
  run cleanly. Commenting out rather than deleting because I don't know
  how Rails behaves when you delete migrations, and so that we preserve
  history of these changes.


brenetic:
👍

PR #511, opened 2016-05-16
Get rid of the weird black vertical grid lines on hits graphs

issyl0:
- There have been weird vertical black lines marking every day on the hits graphs probably due to something weird in the Google Charts API, which makes reading it a problem. This change seems to fix it - it's back to the normal grid lines now. Not sure if it's perfect, but it's certainly easier to read!

Before:

screenshot 2016-05-16 12 32 48

screenshot 2016-05-16 12 39 27

After:

screenshot 2016-05-16 12 46 38

screenshot 2016-05-16 12 39 59

One caveat: this removes the "Transition" line marking the date of Transition on the "all hits" graph, but not the text. Should we remove the text, or try and restore the line somehow?

(cc @fofr)

### DIFF
--- a/app/assets/javascripts/hits.js
+++ b/app/assets/javascripts/hits.js
@@ -44,7 +44,6 @@
                 height: '80%'
               },
               colors: colors,
-              annotations: {style: 'line', stemColor: 'black'},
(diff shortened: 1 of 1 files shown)


issyl0:
Closing in favour of #512 . 👏

PR #512, opened 2016-05-17
Fix black lines on hits graphs

jennyd:
Between 28/04/2016 and 03/05/2016 the hits graphs suddenly gained a black line for every day, instead of only for the transition date:

screen shot 2016-05-17 at 14 23 00

I can't see any relevant changes in the Google Charts release notes [1] - the last documented release was in February, long before we started seeing these problems.

The documentation for chart annotations [2] says that the default value for the annotation role is `''`, which suggests that our existing code should still work, but the examples on that page use `null` for datapoints without annotations. Switching our default value in the helper to `nil` appears to fix the problem, and the annotation for the transition date is shown correctly:

screen shot 2016-05-17 at 14 23 57

[1]: https://developers.google.com/chart/interactive/docs/release_notes#official-releases
[2]: https://developers.google.com/chart/interactive/docs/roles#what-roles-are-available

### DIFF
--- a/app/helpers/hits_helper.rb
+++ b/app/helpers/hits_helper.rb
@@ -54,7 +54,7 @@ module HitsHelper
       rows << {
         c: [
              { v: "Date(#{date.year}, #{date.month - 1}, #{date.day})" },
-             { v: date == transition_date ? 'Transition' : ''},
+             { v: date == transition_date ? 'Transition' : nil },
              *categories.map do |c|
                count_for_category = category_counts[c.name] || 0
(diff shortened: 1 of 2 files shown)


jennyd:
This is an alternative to https://github.com/alphagov/transition/pull/511, which fixes the blank lines but also removes the line for the transition date.

fofr:
:+1:

issyl0:
Your explanation is better than mine, and it preserves the line for Transition date. 👍

PR #514, opened 2016-06-28
Import hits with .jsp extension

jennyd:
We already ignore hits rows which contain various patterns which indicate that we don't need to map them, such as `.js`, but we do want to import hits rows with `.jsp` - this is an extension used by Java Spring apps.

This commit updates the pattern to match either `.js$` or `.js<any non-word character>` so that any `.js` paths with query strings are still matched. This enables us to import `.jsp` hits.

I don't intend to reimport all the hits files after this change has been deployed. There's one site in particular which I know is affected by this (`directgov_local`) so we will reimport those hits files from pre-transition-stats.

I found this problem while working on mappings for that site - we know that we want to map URLs like http://local.direct.gov.uk/LDGRedirect/index.jsp?LGSL=57 but those mappings didn't exist after [creating mappings from host paths](https://github.com/alphagov/transition/blob/master/lib/tasks/import/mappings_from_host_paths.rake).

### DIFF
--- a/lib/transition/import/hits/ignore.rb
+++ b/lib/transition/import/hits/ignore.rb
@@ -33,7 +33,7 @@ module Transition
         PATTERNS = [
           # Generic site furniture
           '.*\.css',
-          '.*\.js',
+          '.*\.js(\W|$)',
           '.*\.gif',
           '.*\.ico',
           '.*\.jpg',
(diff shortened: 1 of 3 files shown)


jennyd:
For reference, this finds the `ImportedHitsFile` rows which we'll need to delete in order to re-import the files after deploying this change:

```
ImportedHitsFile.where('filename LIKE ?', 'data/pre-transition-stats/hits/local.direct.gov.uk%')
```

PR #515, opened 2016-07-01
Fix failing features for graphs

h-lame:
We noticed in #514 that the specs were failing on CI.  While investigating what seemed to be an unrelated change we noticed that the specs were also failing on master.  It seems that something changed in the google charts api that our JS testing infrastructure (poltergeist and phantomjs) didn't like.  This PR upgrades those components and changes how we invoke the charts API.

There's more details in each commit.


jennyd:
Thank yooooouuuuu @h-lame 👻  📉  Tests pass now so I think this is 🍏  but we mostly paired on this so it would be good for someone else to review too.

issyl0:
👻 📉 ⌚ ⭐ 📈

PR #516, opened 2016-07-04
Use capybara matchers for interacting with svg text

h-lame:
Old versions of poltergeist didn't extract the values from svg text
elements correctly so we used a custom matcher that used nokogiri.  The
downside here is that it couldn't take advantage of capybara's ability
to wait a while for the elements to appear (allowing for AJAX or animation
effects to finish changing the page).

Without doing this we would get some build failures that we think were
down to timing problems.  After changing to use the capybara matcher we
ran the specs in a few loops of 100 runs and got no failures.

To address intermittent failures since: #515.


jennyd:
Thanks @h-lame! 🎉 🍏 💯

PR #517, opened 2016-07-27
Ignore malformed URIs

MatMoore:
Some of the hits data has malformed URIs, for example

/statistics/child%E2?%3Fpoverty%E2%3F%3Fstats.htm

instead of

/statistics/child-poverty-stats.htm

This is preventing us from running the transition import:
the Transition_load_all_data jenkins job fails as soon as it tries
to handle one of these paths.

There is already a HostPath entry for the correct path, as well
as the one with the malformed the URI.

This change means we skip associating the HostPath to the Mapping
if the URI cannot be decoded.

### DIFF
--- a/lib/transition/import/hits_mappings_relations.rb
+++ b/lib/transition/import/hits_mappings_relations.rb
@@ -60,7 +60,12 @@ module Transition
         host_paths.includes(:host).find_each do |host_path|
           site = host_path.host.site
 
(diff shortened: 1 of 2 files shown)


sihugh:
We've got a PR coming to fix this in the very near future (we need to bump the optic14n Gem)
https://github.com/alphagov/optic14n/pull/7

sihugh:
Need to keep Bouncer in step with it too

PR #518, opened 2016-07-27
Bump Optic14n to fix query canonicalization

sihugh:
To fix https://errbit.publishing.service.gov.uk/apps/530f56010da115868600173d/problems/577c047a65786352db2b0400

This error occurs when canonicalizing query strings.  It appears that not every
character emerges unscathed from this process.  In this case, having `%E2`
(which is equivalent to "â")  in a key of a query string results in the error
found in Errbit.

See https://trello.com/c/ihTQRRMX/325-encoding-error-on-hits-import-in-transition
for further explanation and context.


sihugh:
Should be merged in conjunction with https://github.com/alphagov/bouncer/pull/138

PR #519, opened 2016-08-05
Govuk Transition signon config

djsd123:
This change will allow transition to use auth tokens defined in hiera when it's deployed.

This should eliminate the need to replace the file each time per environment.


deanwilson:
How do these `ENV` vars get sent to the app? Can we link to that commit please?

alexmuller:
Added the environment variables commit into this commit message.

deanwilson:
Thanks for the added context. Merge away and we can test in integration.

PR #520, opened 2016-08-05
Make the sites table filterable

jennyd:
Several organisations have lots of sites now, which makes it harder to find the one you're looking for in the table. Making the table filterable, as we do for the organisations index, helps with this.

GDS has by far the most sites (214; FCO has 163, BIS 66 and UKTI 58) and the majority of organisations have only one site, so this is very much scratching our own itch. Should we only add the filter header row if there are more than 5 sites, or are we happy for it to be there for everyone?

screen shot 2016-08-05 at 17 41 10

### DIFF
--- a/app/views/organisations/show.html.erb
+++ b/app/views/organisations/show.html.erb
@@ -12,12 +12,20 @@
 
 <% unless @sites.empty? %>
   <h2>Sites</h2>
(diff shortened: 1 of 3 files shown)


h-lame:
Have we ever done any user research on the cognitive load of adding the filter for users without enough sites to need it?  Maybe we could merge it as-is and then run that research?

PR #522, opened 2016-08-18
Fix tag list field to fill with comma separated tags

davidbasalla:
For https://trello.com/c/ZYSIdD8R/466-tags-in-edit-mapping-form-are-mangled-transition-3

Previously the text_field for a tag_list_field would merge multiple tags into one string (ie with multiple tags would be filled with words only separated by a space, e.g. "fee fi of"). On saving the mapping, this would create a new tag. The problem seems to have been that when using the FormHelper text_field method, calling the `tag_list` method did not return a comma separated string. So we explicitly set the value to be the string-ified array.

In order to make testing easier, we replace `acts_as_taggable` with `acts_as_ordered_taggable` which ensures that the tags remain ordered. Due to some interplay between the 'paper-trail' and the 'acts-as-taggable-on' gems, this requires a change to the ignore block for the 'paper_trail' definition in the Mapping model.

# Before (when editing a mapping with multiple tags)
screen_shot_2016-07-08_at_14 31 24

# After
screen_shot_2016-07-08_at_15 58 51

### DIFF
--- a/app/models/mapping.rb
+++ b/app/models/mapping.rb
@@ -12,13 +12,7 @@ class Mapping < ActiveRecord::Base
   SUPPORTED_TYPES = %w(redirect archive unresolved)
 
   acts_as_taggable
-  has_paper_trail :ignore => [:tag_list => Proc.new { |mapping|
-    # tag_list appears to ActiveModel::Dirty/paper_trail to be an
-    # ActsAsTaggableOn::TagList but the original value from the database is a
-    # string. This means that it is "different" even though it isn't.
-    # Comparing the stringified version avoids that problem.
-    mapping.tag_list_was == mapping.tag_list.to_s
-  } ]
+  has_paper_trail
 
   belongs_to :site
   has_many :hits
--- a/features/mapping_tag.feature
+++ b/features/mapping_tag.feature
@@ -11,6 +11,8 @@ Scenario: Adding tags to a mapping
   And I save the mapping
   Then I should see "Mapping saved"
   And the mapping should have the tags "fee, fi, fo"
+  When I edit that mapping
+  Then I should see the tags "fee, fi, fo"
 
 Scenario: Adding tags when bulk adding mappings
   Given I have logged in as a GDS Editor
--- a/app/views/shared/_tag_list_field.html.erb
+++ b/app/views/shared/_tag_list_field.html.erb
@@ -26,7 +26,7 @@
   <div class="row">
     <div class="<%= modal ? 'col-md-12' : 'col-md-8' %>">
       <% if defined? form %>
-        <%= form.text_field :tag_list, class: 'form-control js-tag-list' %>
+        <%= form.text_field :tag_list, class: 'form-control js-tag-list', value: form.object.tag_list.to_s %>
       <% else %>
         <%= text_field_tag :tag_list, tag_list.to_s, class: 'form-control js-tag-list' %>
       <% end %>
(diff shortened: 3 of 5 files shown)


davidbasalla:
I'm slightly suspicious since changing the line in mapping.rb to `has_paper_trail :ignore => [:tag_list]` also makes the tests pass...

h-lame:
This seems weird, but there's a wiki entry on the acts_as_taggable_on github that confirms it: https://github.com/mbleigh/acts-as-taggable-on/wiki/Add-a-tag-field-to-forms-in-Rails--4.2-and-above.  Seems like this should have been in the [UPGRADING](https://github.com/mbleigh/acts-as-taggable-on/blob/master/UPGRADING.md) docs, as it looks like this has been broken since #508 back in March.  Good to have tests now in case the behaviour changes again with another rails or acts_as_taggable_on upgrade.

app/models/mapping.rb
```diff
@@ -11,13 +11,13 @@ class Mapping < ActiveRecord::Base
...
 
-  acts_as_taggable
+  acts_as_ordered_taggable
   has_paper_trail :ignore => [:tag_list => Proc.new { |mapping|
     # tag_list appears to ActiveModel::Dirty/paper_trail to be an
     # ActsAsTaggableOn::TagList but the original value from the database is a
     # string. This means that it is "different" even though it isn't.
     # Comparing the stringified version avoids that problem.
-    mapping.tag_list_was == mapping.tag_list.to_s
+    mapping.tag_list_was == mapping.tag_list
```
h-lame: You probably want to leave this as it was.  As the comment says, `mapping.tag_list_was` is a string, but `mapping.tag_list` is a `ActsAsTaggableOn::TagList` object so this comparison will always fail unless you stringify the tag_list.

h-lame:
Suspect that no tests failed because we must not be testing the paper_trail integration stuff.  Maybe it's worth writing something that does, because that tag_list thing is bound to catch someone at some point.

app/models/mapping.rb
davidbasalla: When I switched to `acts_as_ordered_taggable`, I had to update this to get the tests to pass, without fully understanding why. Will investigate more! writing more tests around this sounds good too

h-lame: Probably because the tags are now ordered and they weren't before so you'd get natural db order when you asked for tag_list rather than the order you specified when setting them.

h-lame: Maybe we shouldn't use acts_as_ordered_taggable if it makes testing easier, but we have to change a bunch of tests to make them pass again?

davidbasalla:
Have reworked this PR a little bit. There's an extra commit now removing the ignore block as it's not needed since we upgraded the 'acts-as-taggable-on' gem.

app/models/mapping.rb
```diff
@@ -12,13 +12,7 @@ class Mapping < ActiveRecord::Base
   SUPPORTED_TYPES = %w(redirect archive unresolved)
 
   acts_as_taggable
-  has_paper_trail :ignore => [:tag_list => Proc.new { |mapping|
-    # tag_list appears to ActiveModel::Dirty/paper_trail to be an
-    # ActsAsTaggableOn::TagList but the original value from the database is a
-    # string. This means that it is "different" even though it isn't.
-    # Comparing the stringified version avoids that problem.
-    mapping.tag_list_was == mapping.tag_list.to_s
-  } ]
```
h-lame: :+1: - good sleuthing!

features/mapping_tag.feature
```diff
@@ -11,6 +11,8 @@ Scenario: Adding tags to a mapping
   And I save the mapping
   Then I should see "Mapping saved"
   And the mapping should have the tags "fee, fi, fo"
+  When I edit that mapping
+  And I should see the tags "fee, fi, fo"
```
jennyd: Minor point: this should be a `Then` step, not an `And` (`When`)

PR #524, opened 2016-09-20
Opt in to new gds api adapters behaviour

h-lame:
For: https://trello.com/c/B5z6Dq8q/495-get-our-apps-up-to-date-on-new-api-adapters-behaviour-update-at-stand-up

Also took the opportunity to add linting to the app and fix all extant errors.  Which means you want to look at this commit by commit (and probably skim-read the massive "Appease linter" commit).  As part of adding the linter I modernised the jenkins.sh script too, which probably means we'll want to update jenkins branch builder to call jenkins.sh, not jenkins_branches.sh (which is deleted).


lib/transition/import/hits.rb
```diff
@@ -64,9 +64,9 @@ def self.from_tsv!(filename)
 
           Hit.transaction do
             import_record = ImportedHitsFile.where(
-                filename: relative_filename).first_or_initialize
+              filename: relative_filename).first_or_initialize
 
-            job.skip! and next if import_record.same_on_disk?
+            job.skip! && next if import_record.same_on_disk?
```
h-lame: Seems we can mutate the cop to only look at conditionals (see https://github.com/bbatsov/rubocop/blob/bcb42a2e5960c56ef0a1c8b4b4093ae5b3e71ce8/config/default.yml#L185-L191).  I'll make that change locally, revert, and re-run see what happens.

lib/transition/import/csv/import_batch_row.rb
```diff
@@ -23,7 +23,7 @@ def data_row?
 
         def type
           @_type ||= case
-                     when new_value && ((new_value.upcase == 'TNA') || new_url_is_a_national_archives_url?) then 'archive'
+                     when new_value && (new_value.casecmp('TNA').zero? || new_url_is_a_national_archives_url?) then 'archive'
```
h-lame: Ugh, yeah, this is weird.  I'll see what I can do.

app/models/organisation.rb
```diff
@@ -33,56 +32,59 @@ class Organisation < ActiveRecord::Base
...
-               FROM   daily_hit_totals
-                      INNER JOIN hosts
-                              ON hosts.id = daily_hit_totals.host_id
-               WHERE  daily_hit_totals.http_status = '404'
-               AND daily_hit_totals.total_on >= (current_date - 30)
-               GROUP BY site_id) AS error_counts ON error_counts.site_id = sites.id
-    postgreSQL
-  ).group('organisations.id').order('error_count DESC NULLS LAST') }
+  scope :leaderboard, -> {
+    select(%{
```
h-lame: That one was actually me, the linter messed up the indentation so I "fixed" it.  I'll move back to a heredoc.

app/models/mapping.rb
```diff
@@ -141,17 +134,26 @@ def self.ensure_url(uri)
...
       self.path = url.request_uri
     end
   rescue Addressable::URI::InvalidURIError
     # The path isn't parseable, so leave it intact for validations to report
+    Rails.logger.warn("Unparseable URI #{path} in mapping")
   end
 
   def canonicalize_path
-    self.path = site.canonical_path(path) unless (site.nil? || path == '/' || path =~ /^[^\/]/)
+    self.path = site.canonical_path(path) unless site.nil? || path == '/' || path =~ /^[^\/]/
```
h-lame: Probably the thing to do is extract the conditionals to a method, so I'll do that.

app/helpers/mappings_helper.rb
```diff
@@ -14,16 +13,17 @@ module MappingsHelper
...
+        href = title_link[1]
         active            = options[:active] == title
         html_opts         = {}
         html_opts[:class] = 'active' if active
 
         result << content_tag(:li, html_opts) do
           link_to(title, active ? '#' : href)
         end
-      end.html_safe
+      }.html_safe
```
h-lame: Yeah, there's a cop about not chaining on `do...end` blocks, and there were a few of these to fix up.  I don't know if there's an alternative way - but I'll see if anything can be re-arranged.

lib/transition/import/dns_details.rb
```diff
@@ -18,14 +18,18 @@ def import!
         hosts.each do |host|
           $stderr.print '.'
 
-          if cname_record = record(host.hostname, Resolv::DNS::Resource::IN::CNAME)
+          cname_record = record(host.hostname, Resolv::DNS::Resource::IN::CNAME)
```
h-lame: It was complaining about assignment in the conditional as it's often a typo of / confused for equality.  I'll see if there's a different way to arrange this.

lib/transition/import/csv/import_batch_row.rb
h-lame: Just disabling this cop because this isn't as obvious at all.

app/models/user.rb
```diff
@@ -25,7 +25,7 @@ def can_edit_site?(site_to_edit)
         own_organisation == site_to_edit.organisation ||
         site_to_edit.organisation.parent_organisations.include?(own_organisation) ||
         site_to_edit.extra_organisations.include?(own_organisation) &&
-        site_to_edit.global_type.blank?
+          site_to_edit.global_type.blank?
```
h-lame: ok, so the reason for this is that what we have is: 

```ruby
a || b || c || d && e
```

What this actually means is:

```ruby
a || b || c || (d && e)
```

So rubocop sees the `d && e` part as a new expression that should be indented separately.  I wonder if that's actually what was intended, or if we meant `(a || b || c || d) && e`.

jennyd: Aha! The intent is `(a || b || c || d) && e` - `e` is only about the site whereas all the others are about the privileges of the user in relation to the site.

In practice it isn't possible to navigate to a page to add mappings for a globally archived/redirected site, but if you know the URL then it is possible to create mappings for them - I've just done it in integration for directgov_localadmin :-/ Those mappings won't be used for a site with a global status, but it shouldn't be possible. We should check that fixing the logic here fixes that bug.

I don't think this is a security risk because it still restricts users to editing mappings for sites they should be allowed to edit. Any mappings created via this bug would only be used by Bouncer if we removed the global status from the site, in which case we would have to sort out the mappings in advance anyway (I don't think we've ever done this).

app/helpers/mappings_helper.rb
h-lame: Two options here I think:

```ruby
# option 1.
content_tag :ul, class: 'nav nav-tabs' do
  content = titles_to_links.inject('') do |result, title_link|
    # all the code in the block
  end
  content.html_safe
end

# option 2
content_tag :ul, class: 'nav nav-tabs' do
  titles_to_links.inject('') { |result, title_link| 
    result << build_title_link(title_link) 
  }.html_safe
end

def build_title_link(title_link)
  # all the code in the block
end
```

Any preference?

jennyd: It feels big enough to make it worth the extra method, so I'd go for option 2.

h-lame:
Just rebased some changes in, I think I've resolved everything you picked up @jennyd.

h-lame:
Added a commit to shared global site checking behaviour between mappings controller and batches controllers.

davidbasalla:
Looks great, good job catching those bugs as well. At a later point, we should probably consolidate how the linter is run both on jenkins and in the rake command, but this looks to be happening in other apps too (LocalLinksManager, SpecialistPublisherRebuild). This looks good though, so will merge!

PR #526, opened 2016-09-30
Upgrade gds-api-adapters to fix Whitehall orgs import

jennyd:
The import is failing with a NoMethodError from a ListResponse in api
adapters. It looks like the fix in
https://github.com/alphagov/gds-api-adapters/pull/582 will help this.


jennyd:
I've deployed this to Integration and run the import and it [runs without errors](https://deploy.integration.publishing.service.gov.uk/job/Transition_load_site_config/39/).

davidbasalla:
👍

PR #527, opened 2016-09-30
Add a spec to cover WhitehallOrgs#load_orgs_from_api

h-lame:
In #526 we bumped gds-api-adapters to get a fix for list responses from
the api.  We only spotted the failure when trying to run the whitehall
orgs import in production because our existing specs have a different
codepath that uses a fixture instead of the API.  We add a spec that does
use the API (albeit stubbed via the gds_api/test_helpers) which should
protect us against similar failures when the API changes in the future.


davidbasalla:
Nice! 😎  Looks good to me

sihugh:
👍

PR #528, opened 2016-11-02
Strip whitespace from the URL in MappingsController#find_global

issyl0:
- This caused me problems when copying and pasting from a spreadsheet
  into the "go to site or mapping box" - I pasted something with
  trailing whitespace and it didn't go anywhere.
- So, strip both leading and trailing whitespace from the
  `params[:url]` field before processing it further.

### DIFF
--- a/app/controllers/mappings_controller.rb
+++ b/app/controllers/mappings_controller.rb
@@ -89,10 +89,13 @@ class MappingsController < ApplicationController
     # This allows finding a mapping without knowing the site first.
     render_error(400) and return unless params[:url].present?
 
-    url = if !Transition::PathOrURL.starts_with_http_scheme?(params[:url])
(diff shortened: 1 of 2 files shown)


app/controllers/mappings_controller.rb
```diff
@@ -89,6 +89,9 @@ def find_global
     # This allows finding a mapping without knowing the site first.
     render_error(400) and return unless params[:url].present?
 
+    # Strip leading and trailing whitespace before any processing.
+    params[:url] = params[:url].strip
```
jennyd: Could we do this without modifying `params`? I think we tried hard to avoid that elsewhere in Transition. Otherwise 👍

issyl0: Sure. Is it better now? I've made `stripped_url`.

jennyd:
👍

PR #530, opened 2016-11-11
Change column types on mappings_batch_entries to match mappings

jennyd:
In particular because `archive_url` was too small for National Archives URLs which are usually long. Batch entries are the precursor to mappings so their column types should match where possible, so that users can create mappings via batches with the same constraints as when doing them individually.

Also use a long `archive_url` in a test which fails without this migration.

See Errbit for [examples of data which trigger this problem](https://errbit.publishing.service.gov.uk/apps/530f56010da115868600173d/problems/5825f6646578633fd99a0b00) - the test URL comes from this data.

There's some context for the large schema diff in the first commit message.

https://trello.com/c/MGtcYgQV/549-fix-limit-on-archive-url-length-when-importing-batches-in-transition


jennyd:
I've updated the db structure in Bouncer to match: https://github.com/alphagov/bouncer/pull/141

davidbasalla:
Looks good, merging!

PR #533, opened 2017-01-23
Add new Fastly IP addresses

alexmuller:
Fastly have set up a service map for Bouncer and have allocated us new
IP address for that map.

We should be moving things away from the other 2 methods of redirection
and should be using just this one map.

### DIFF
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -15,9 +15,10 @@ class Host < ActiveRecord::Base
 
   scope :excluding_aka, -> { where(canonical_host_id: nil) }
 
-  FASTLY_ANYCAST_IPS = ['23.235.33.144', '23.235.37.144'].freeze # To be used for new root domains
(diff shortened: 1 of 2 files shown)


alexmuller:
```
One or more errors were raised in the Javascript code on the page.

ReferenceError: Can't find variable: gvjs_U
```

jennyd:
This change looks fine to me, but it would be good to add a [test](https://github.com/alphagov/transition/blob/788dfbc9ba6e10f97e8a468c3ffb3ede84b91fe2/spec/models/host_spec.rb#L132-L177) for the bouncer CNAME.

The unrelated test failures are annoying - we've had intermittent test failures around the google charts js before. When I run the tests locally I get between 0 and 2 of these errors (the 3 CI builds have 7, 6, 7 of them) - I don't get any js errors in the browser either locally or in Production. Haven't tried to work out what's going on, and I'm unlikely to have time to investigate any time soon...

sihugh:
@jennyd 👀  please?

PR #534, opened 2017-02-06
Suppress javascript errors in Poltergeist

sihugh:
We've got intermittent-ish test failures around the charts js in Transition
again, which get in the way of us making even small unrelated changes to the
app. This time it's variations on this in the hits_summary feature.

We're disabling raising of javascript errors in Poltergeist so that we can get
builds running again.  We've got a [card for investigating the issue further][1]
in our backlog.

Trello story: https://trello.com/c/nvjibWTy/633-transition-test-failures

[1]: https://trello.com/c/kaQn2aYB/528-transition-test-failures

### DIFF
--- a/features/support/poltergeist.rb
+++ b/features/support/poltergeist.rb
@@ -5,7 +5,8 @@ Capybara.register_driver :poltergeist do |app|
   Capybara::Poltergeist::Driver.new(app, {
     phantomjs: Phantomjs.path,
     window_size: [1366, 768],
-    debug: false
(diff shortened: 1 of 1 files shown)


sihugh:
@alexmuller this should hopefully let us build https://github.com/alphagov/transition/pull/533

jennyd:
LGTM - the steps in the features testing the hits graphs are defined usefully enough that they should fail anyway if the graph really is broken.

PR #535, opened 2017-05-10
Remove Dyn DNS redirect IP addresses

alexmuller:
In the very old days we used a Dyn HTTP redirect service for direct.gov.uk and businesslink.gov.uk. We've now changed those redirects to use Bouncer.

There are only 2 domains left in Transition which use these IP addresses. They are not actively redirecting so they need to be removed before this is merged and deployed.

### DIFF
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -17,8 +17,7 @@ class Host < ActiveRecord::Base
 
   FASTLY_BOUNCER_SERVICE_MAP = %w(151.101.2.30 151.101.66.30 151.101.130.30 151.101.194.30).freeze # bouncer.gds.map.fastly.net.
   FASTLY_ANYCAST_IPS = ['23.235.33.144', '23.235.37.144'].freeze # FIXME: These IPs are deprecated, Fastly would like to reallocate them
-  DYN_DNS_IPS        = ['216.146.46.10', '216.146.46.11'].freeze # FIXME: These IPs are Dyn's redirect service, replace them with Fastly
-  REDIRECTOR_IPS     = FASTLY_ANYCAST_IPS + FASTLY_BOUNCER_SERVICE_MAP + DYN_DNS_IPS
+  REDIRECTOR_IPS     = FASTLY_ANYCAST_IPS + FASTLY_BOUNCER_SERVICE_MAP
 
   REDIRECTOR_CNAME = /^(redirector|bouncer)-cdn[^.]*\.production\.govuk\.service\.gov\.uk$/
 


stephenharker:
The two remaining domains are:

contact-ukba.homeoffice.gov.uk
report-ukba.homeoffice.gov.uk

which are in use and being redirected by DynDNS and are not pointing at Bouncer. These two redirects can be removed from Transition and then this PR merged.

tijmenb:
This is still the case so we can't merge this yet:

```rb
irb(main):004:0> y Host.where(ip_address: ['216.146.46.10', '216.146.46.11']).as_json
---
- id: 1745
  site_id: 2
  hostname: report-ukba.homeoffice.gov.uk
  ttl: 48
  cname: 
  live_cname: 
  created_at: !ruby/object:ActiveSupport::TimeWithZone
    utc: 2014-04-09 11:02:00.000000000 Z
    zone: &1 !ruby/object:ActiveSupport::TimeZone
      name: Europe/London
    time: 2014-04-09 12:02:00.000000000 Z
  updated_at: !ruby/object:ActiveSupport::TimeWithZone
    utc: 2017-09-01 08:03:36.723757000 Z
    zone: *1
    time: 2017-09-01 09:03:36.723757000 Z
  ip_address: 216.146.46.10
  canonical_host_id: 
- id: 1743
  site_id: 2
  hostname: contact-ukba.homeoffice.gov.uk
  ttl: 59
  cname: 
  live_cname: 
  created_at: !ruby/object:ActiveSupport::TimeWithZone
    utc: 2014-04-09 11:01:59.000000000 Z
    zone: *1
    time: 2014-04-09 12:01:59.000000000 Z
  updated_at: !ruby/object:ActiveSupport::TimeWithZone
    utc: 2017-09-01 08:04:46.707334000 Z
    zone: *1
    time: 2017-09-01 09:04:46.707334000 Z
  ip_address: 216.146.46.10
  canonical_host_id: 
```

cbaines:
So, I deployed https://github.com/alphagov/transition-config/commit/9ce925400e42c6003bad2140847184c83764826f removing the aliases in question, but that didn't to anything, so I've now deleted the two hostnames, with the reasoning that it's probably better to keep the Transition database in sync with the configuration in the transition-config repository.

```ruby
irb(main):002:0> Host.where(hostname: ['report-ukba.homeoffice.gov.uk', 'contact-ukba.homeoffice.gov.uk']).delete_all
=> 2
```

it's unclear to me why this is blocked on these hosts existing, but anyway, they don't now, so I think this is good to be merged.

```ruby
irb(main):004:0> Host.where(ip_address: ['216.146.46.10', '216.146.46.11']).count
=> 0
```

jennyd:
The aka domains for these two hosts should also be deleted as part of the cleanup for this:

```
aka-report-ukba.homeoffice.gov.uk
aka-contact-ukba.homeoffice.gov.uk
```

If there were ever any hits or host paths associated with any of these domains then that complicates things, but there shouldn't be any if they've never sent any traffic to Bouncer.

It would also be fine to leave all four of these hosts in the database and merge this PR - the only change would be that Transition would no longer think that we redirect these domains, which is fine.

cbaines:
> The aka domains for these two hosts should also be deleted as part of the cleanup for this:
> 
> aka-report-ukba.homeoffice.gov.uk
> aka-contact-ukba.homeoffice.gov.uk

Ok, I've now deleted the Hosts with these hostnames:

```ruby
irb(main):001:0> Host.where(hostname: ['aka-report-ukba.homeoffice.gov.uk', 'aka-contact-ukba.homeoffice.gov.uk']).count
=> 2
irb(main):002:0> Host.where(hostname: ['aka-report-ukba.homeoffice.gov.uk', 'aka-contact-ukba.homeoffice.gov.uk']).delete_all
=> 2
irb(main):003:0> Host.where(hostname: ['aka-report-ukba.homeoffice.gov.uk', 'aka-contact-ukba.homeoffice.gov.uk']).count
=> 0
```

> If there were ever any hits or host paths associated with any of these domains then that complicates things, but there shouldn't be any if they've never sent any traffic to Bouncer.
> 
> It would also be fine to leave all four of these hosts in the database and merge this PR - the only change would be that Transition would no longer think that we redirect these domains, which is fine.

That makes sense. However, as I've already started messing with things, I think it makes more sense to continue now, rather than to stop or backtrack.

cbaines:
So, I noticed when I got in this morning that the Transition_load_all_data Jenkins job had failed on Wednesday morning, and failed again this morning (Thursday).

I've now investigated this, and it looks like the problem was that despite having no foreign key constraints in the entire database, the Transition_load_all_data job does actually depend on the Host with the hostname contact-ukba.homeoffice.gov.uk existing, as its referenced by 
167 host_paths [1].

This was my mistake, as I should have at least checked to see if there were any foreign key constraints.

1: https://github.com/alphagov/govuk-org-notes/blob/master/Debugging%20the%20Transition_load_all_data%20Jenkins%20job%20failures.org#how-many-host_paths-are-broken

> If there were ever any hits or host paths associated with any of these domains then that complicates things, but there shouldn't be any if they've never sent any traffic to Bouncer.

As above, turns out this is an issue, at least for the contact related hostnames.

> It would also be fine to leave all four of these hosts in the database and merge this PR - the only change would be that Transition would no longer think that we redirect these domains, which is fine.

As touching the database has broken things, and there are no foreign key constraints but obviously places in transition/bouncer that assume referential integrity, I think this is the best option. I've now downloaded some historic data to work out what was deleted, and restored the 4 deleted host records (full details in [2]).

I'm still unsure how the transition-config relates to the configuration of transition, but I'm going to suggest reverting the transition-config change as well.

I'm uncertain where to go from here. The original task I set out to solve was to advise infrastructure on tidying up Dyn redirects. As it happens, the resolution to this was to realise that there are no Dyn redirects to tidy up, as they are not within the control of GOV.UK.

I'm not sure what to do with this PR and I'm hesitant to do anything, as I think I've already done enough damage fumbling around with no specific intent, and its not clear to me what the intent behind this change is.

2: https://github.com/alphagov/govuk-org-notes/blob/master/Debugging%20the%20Transition_load_all_data%20Jenkins%20job%20failures.org

alexmuller:
I think you're doing ok! I should've explained this change more clearly in the first place :)

The bit @jennyd mentioned above is what I should have reinforced when I originally opened the PR:

> It would also be fine to leave all four of these hosts in the database and merge this PR - the only change would be that Transition would no longer think that we redirect these domains, which is fine.

The only thing this pull request affects is certain parts of the transition interface that depend on the `Host.redirected_by_gds?` method.

My gut feeling would have been to push forward and remove the host paths from the database. In my opinion this pull request is technically correct and good to merge. `Host.redirected_by_gds?` should indicate if a host is redirected by a service that GDS controls, and these Dyn IP addresses are not controlled by GDS. But I'm pretty sure I don't get a vote anymore :)

PR #536, opened 2017-05-23
Upgrade to rails 5.1.1

Rosa-Fox:
We have upgraded Transition to use rails 5.1.0

Trello: https://trello.com/c/XQgx7sdX/86-upgrade-transition-to-a-supported-version-of-rails

We will upgrade `paper_trail` separately: https://trello.com/c/Uyo9qshv/156-upgrade-papertrail-gem-in-transition

We can't fix the following deprecation warning because it is still part of another gem (we suspect either `paper_trail` or `acts-as-taggable`)

>DEPRECATION WARNING: The behavior of `changed_attributes` inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). To maintain the current behavior, use `saved_changes.transform_values(&:first)` instead.

- https://github.com/alphagov/bouncer/pull/143

Thank you


app/controllers/organisations_controller.rb
```diff
@@ -8,6 +8,6 @@ def show
     @organisation = Organisation.find_by_whitehall_slug!(params[:id])
     sites = @organisation.sites.includes(:hosts)
     extra_sites = @organisation.extra_sites.includes(:hosts)
-    @sites = sites.concat(extra_sites).sort_by!(&:abbr)
+    @sites = sites + extra_sites.sort(&:abbr)
```
carvil: Is this sorting the sites as expected? Before we were concatenating the 2 lists before sorting them, but I'm not sure if that's what's happening now. It looks like it's sorting `extra_sites` and then adding sites to it, is that correct?

app/models/organisation.rb
```diff
@@ -1,6 +1,4 @@
 class Organisation < ActiveRecord::Base
-  belongs_to :parent, class_name: Organisation, foreign_key: 'parent_id'
```
carvil: What was the reason for deleting this `belongs_to` relation?

bin/yarn
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env ruby
+VENDOR_PATH = File.expand_path('..', __dir__)
```
carvil: I don't think this file is needed, could you please delete it?

config/application.rb
```diff
@@ -1,10 +1,12 @@
...
 
 # Pick the frameworks you want:
-require "active_record/railtie"
-require "action_controller/railtie"
-require "action_mailer/railtie"
-require "sprockets/railtie"
+# require "active_record/railtie"
+# require "action_controller/railtie"
+# require "action_mailer/railtie"
+# require "sprockets/railtie"
```
carvil: Could you please delete these commented out require statements?

config/application.rb
```diff
@@ -33,8 +38,5 @@ class Application < Rails::Application
     # We should consider changing how we do this.
     # Route exceptions to our custom error pages.
     config.exceptions_app = self.routes
-
-    # Do not swallow errors in after_commit/after_rollback callbacks.
-    config.active_record.raise_in_transactional_callbacks = true
```
carvil: Is this the default now?

config/cable.yml
```diff
@@ -0,0 +1,9 @@
+development:
```
carvil: We don't need action cable, could you please delete this file?

config/initializers/application_controller_renderer.rb
```diff
@@ -0,0 +1,6 @@
+# Be sure to restart your server when you modify this file.
+
```
carvil: I think it's ok to delete this file too

config/initializers/cookies_serializer.rb
```diff
@@ -1,3 +1,5 @@
 # Be sure to restart your server when you modify this file.
 
-Rails.application.config.action_dispatch.cookies_serializer = :hybrid
+# Specify a serializer for the signed and encrypted cookie jars.
+# Valid options are :json, :marshal, and :hybrid.
+Rails.application.config.action_dispatch.cookies_serializer = :json
```
carvil: Was there a reason for changing from `hybrid` into `json`?

config/puma.rb
```diff
@@ -0,0 +1,47 @@
+# Puma can serve each request in a thread from an internal thread pool.
+# The `threads` method setting takes two numbers a minimum and maximum.
+# Any libraries that use thread pools should be configured to match
```
carvil: We won't be using Puma, which means we can delete this file too.

features/support/helpers/factory_girl.rb
```diff
@@ -1,3 +1,8 @@
 require 'factory_girl'
 
 World(FactoryGirl::Syntax::Methods)
+
+
+Given(/^I am in a debugging mode$/) do
+  ENV['TESTING'] = 'true'
```
carvil: Is this needed, or was just for debugging purposes?

app/models/organisation.rb
Davidslv: Organisation was deleted and this was a leftover

Davidslv: https://github.com/alphagov/transition/pull/536/commits/62642815648fe91769828794c34dc86a3f348b40

features/support/helpers/factory_girl.rb
Davidslv: This was just for debugging purposes - will remove it.

app/models/organisation.rb
carvil: Ah, ok, I reviewed the overall PR rather than individual commits and missed that!

config/application.rb
Davidslv: no longer valid https://stackoverflow.com/questions/28006358/undefined-method-raise-in-transactional-callbacks-for-activerecordbaseclass

carvil:
@Davidslv looks like the tests are failing. It's a bit hard to see why, though. What do you think of addressing those deprecation warnings to have a clearer output?

```
DEPRECATION WARNING: The behavior of `attribute_changed?` inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). To maintain the current behavior, use `saved_change_to_attribute?` instead. (called from update at /var/lib/jenkins/workspace/sition_upgrade-rails-take-2-MPDI727S5XXM5HQT34EV7GEJOJZR3GFBQGOJUG5X53RUFRQTK74A/app/controllers/sites_controller.rb:9)
```

Davidslv:
@carvil I don't think I can address the warning, it's quite annoying that they send the deprecation warning (rails) but its them using the method 

```
ruby/gems/2.3.0/gems/activemodel-5.1.0/lib/active_model/dirty.rb: def changed_attributes
```

Going to check why is it breaking again - i though it was green on my local machine... must have missed something

Gemfile
```diff
@@ -1,59 +1,59 @@
...
-gem 'pg', '0.18.4'
-gem 'optic14n', '2.0.1'     # Ideally version should be synced with bouncer
-gem 'gds-sso', '12.0.0'
-gem 'govuk_admin_template', '4.2.0'
-gem 'plek', '1.12.0'
-gem 'htmlentities', '4.3.4'
-gem 'kaminari', '0.16.3'
+gem 'rails', '5.1.0'
+gem 'activerecord-session_store'
+gem 'unicorn'#, '4.6.2'
```
carvil: Before you squash all these commits, could you please remove the commented out versions?

Gemfile
```diff
@@ -1,59 +1,59 @@
...
-gem 'activerecord-import', '0.12.0'
-gem 'sidekiq', '4.1.1'
-gem 'redis-namespace', '1.5.2'
+gem 'activerecord-import'
+gem 'sidekiq'
+gem 'redis-namespace'
 
 # We use Errbit for tracking exceptions, which needs the airbrake gem. Config
 # for Errbit is in alphagov-deployment.
-gem 'airbrake', '~> 4.3.0'
```
h-lame: Airbrake 5 breaks deploy notifications in our errbit.  We've a branch of 4.3 that fixes dep warnings in rails 5 though. 
 See https://docs.publishing.service.gov.uk/manual/upgrade-rails.html#airbrake for more details.

Davidslv: That explains why there wasn't any errors appearing, thanks - will address this

Davidslv:
@carvil I've now tested and fixed the problems in integration.

PR #540, opened 2017-08-31
Only expose hosts with a CNAME or IP address

cbaines:
https://trello.com/c/6F0cfpiB/14-%F0%9F%94%A5bouncer-limits-in-fastly%F0%9F%94%A5


cbaines:
@jennyd @sihugh @issyl0 or anyone else familiar with Transition and Bouncer, it would be good to get your opinions on this proposed change?

tijmenb:
For the record, these are all the hosts that will be removed from Fastly after this change:

https://gist.github.com/tijmenb/93c78b29f273ca7537dca8fe8b90b3b3

PR #611, opened 2018-09-20
Fix or ignore brakeman warnings

barrucadu:


bevanloon:
Would it be useful to have a quick overview of why it's safe to ignore these? - there are quite a few of them in the commit and they are a mix of SQLi and CSRF errors. Without delving in to the code myself, I've got no real pointers as to why we think they are safe to ignore in the first instance.

PR #615, opened 2018-10-29
Hard-code the external api/organisations for config

schmie:
- The Jenkins jobs loading the config for transition are currently
  talking to whitehall-admin.internal/api/organisations, while the api
  is exposed via www.gov.uk/api/organisations as well

- During AWS migration, internal access to whitelladmin is temporary
  unavailable so we need this change for the config updates to work

- After AWS migration I still do not find the idea of keeping it too
  bad, because it is a test of the external facing API if anything...

solo: @schmie


schmie:
Superseeded by #616

PR #678, opened 2019-03-15
Add a description for import:orgs_sites_hosts task

jennyd:
This means that it shows up in the list of rake tasks. This task is useful
if you want to import a single site config file repeatedly as you're working
on it.


cbaines:
(I've kicked off the CI process now)

jennyd:
Thanks Chris and Bevan! 🎉 💖

PR #686, opened 2019-04-03
Add new Bouncer IP addresses

jennyd:
Transition needs to know the IP addresses and hostnames used by Bouncer in order to correctly report the transition status of sites and hosts.

@timblair asked us to use these IP addresses if we weren't able to use CNAME, ANAME or ALIAS records, rather than the anycast IPs.

Also tweak the tests for `redirected_by_gds?` to cover these cases better.

### DIFF
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -20,8 +20,9 @@ class Host < ActiveRecord::Base
   }
 
   FASTLY_BOUNCER_SERVICE_MAP = %w(151.101.2.30 151.101.66.30 151.101.130.30 151.101.194.30).freeze # bouncer.gds.map.fastly.net.
+  FASTLY_NEW_BOUNCER_IPS = %w(151.101.0.204 151.101.64.204 151.101.128.204 151.101.192.204).freeze
   FASTLY_ANYCAST_IPS = ['23.235.33.144', '23.235.37.144'].freeze # FIXME: These IPs are deprecated, Fastly would like to reallocate them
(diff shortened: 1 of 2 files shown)


app/models/host.rb
```diff
@@ -20,8 +20,9 @@ class Host < ActiveRecord::Base
   }
 
   FASTLY_BOUNCER_SERVICE_MAP = %w(151.101.2.30 151.101.66.30 151.101.130.30 151.101.194.30).freeze # bouncer.gds.map.fastly.net.
+  FASTLY_NEW_BOUNCER_IPS = %w(151.101.0.204 151.101.64.204 151.101.128.204 151.101.192.204).freeze
```
jennyd: I'm happy to rename `FASTLY_NEW_BOUNCER_IPS` if you can suggest something more accurate - I don't know what these ones should actually be called :)

timblair: I think that's fine for now. I have no better thing to call them!

timblair:
@seb-szy Can you double-check the new IPs here, please?

seb-szy:
Hi,

IP addresses added to a config match these recommended by Fastly.

Seb

timblair:
Checked in staging. Before:

transition-ip-updates-before

After:

transition-ip-updates-after

Deploying to production now.

timblair:
Now done in production. You are officially "Live" 🎉 Thanks for the assistance, @jennyd.

jennyd:
Amazing, thanks! 💖

PR #757, opened 2019-07-22
 Allow site to be nil while importing

thomasleese:
We're seeing [a problem while running the import that the `site` is `nil`](https://deploy.blue.production.govuk.digital/job/transition_import_hits/31/console).

```
Adding missing mapping_id/canonical_path to host paths ... 
rake aborted!
NoMethodError: undefined method `site' for nil:NilClass
/data/vhost/transition/releases/20190715125102/lib/transition/import/hits_mappings_relations.rb:65:in `block in connect_mappings_to_host_paths!'
```

Instead of failing the entire import, we should warn that this is the case.

### DIFF
--- a/lib/transition/import/hits_mappings_relations.rb
+++ b/lib/transition/import/hits_mappings_relations.rb
@@ -67,13 +67,13 @@ module Transition
           canonical_path = site.canonical_path(host_path.path)
           mapping_id = Mapping.where(
             path: canonical_path, site_id: site.id
-).pluck(:id).first
+          ).pluck(:id).first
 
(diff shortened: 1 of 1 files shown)


benthorner:
Looking in the DB, I can't find any hosts without a site, so I'm wondering why this change is necessary?

benthorner:
I think we might want to revise the description to be that the *host* is nil.

thomasleese:
Yeah, I couldn't work it out either, hopefully once I've run the task with this change, we'll be able to see why it was failing.

PR #820, opened 2019-11-06
Reach into paper trail to display originator

bevanloon:
Navigating to the "Redirection Whitelist" in Transition Admin is error in the admin application with "We're sorry something went wrong". This is reflected in Sentry by a "ActionView::Template::Error: undefined method `originator'"

The WhitelistedHost model does not contain an `originator` field.
Instead this is exposed by `paper_trail` (a gem that provides version
history on a model) which the model has referenced.

Update the view so we refer to `originator` through `paper_trail` rather
than directly on the model. This should fix the error we are seeing.


issyl0:
I was going to ask the same thing. This probably came about due to #643 where we fixed most of the failing tests but didn't do enough manual testing.

bevanloon:
@thomasleese @issyl0 - added a test if you'd like to have another look please

PR #833, opened 2019-11-29
Enable rails-specific rules for rubocop-govuk

benjamineskola:
trello: https://trello.com/c/Lijki8aj/1584-3-enable-rails-rubocop-rules-via-rubocop-govuk-on-all-our-rails-apps

### DIFF
--- a/app/helpers/versions_helper.rb
+++ b/app/helpers/versions_helper.rb
@@ -2,7 +2,7 @@
 
 module VersionsHelper
   def value_or_blank(value)
-    value.blank? ? "<blank>" : value
+    value.presence || "<blank>"
   end
 
   def friendly_changeset_title_for_type(value)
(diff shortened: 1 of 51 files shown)


app/helpers/versions_helper.rb
```diff
@@ -2,7 +2,7 @@
 
 module VersionsHelper
   def value_or_blank(value)
-    value.blank? ? "<blank>" : value
+    value.presence || "<blank>"
```
AlanGabbianelli: Nice!

benjamineskola:
@AlanGabbianelli There already is an `ActionController`, which Rubocop wants the controllers to inherit from (and most of them do). These two seem to be a special case, for reasons that aren't clear to me, and making them inherit from `ActionController` causes the tests to fail (because it enables authentication for them, for example).

There might be a better way of handling the inheritance, though. I'll add this information when I squash the commits.

PR #866, opened 2020-01-17
Fix auth0 login

lawrence-dxw:


lawrence-dxw:
Oops, wrong repository, sorry!

PR #877, opened 2020-02-13
Fix date specific test

sihugh:
2014 doesn't appear in the select list once the date rolls around to 2020.


sihugh:
I think we're just old

PR #894, opened 2020-03-06
Fix gem deprecation warnings

issyl0:
- See commits for more detail on the specific deprecations.
- The reason for this is that ahead of upgrading to Rails 6, I ran the
  tests locally to make sure my Docker setup worked. I noticed some of
  these which will only get lost in all of the Rails upgrade related
  deprecations, so fix them first then Dependabot can rebase.


issyl0:
The other one in the output is:

```
Calling `DatabaseCleaner.connections` is deprecated, and will be removed in database_cleaner 2.0. Use `DatabaseCleaner.cleaners`, instead.
```

But I can't find where that comes from - nothing in Transition's code uses `DatabaseCleaner.connections`.

PR #1000, opened 2020-09-22
Use SimpleCov to monitor code coverage

benthorner:
https://trello.com/c/epNWudCR/176-write-an-rfc-for-continuous-deployment-and-seek-review

We should always monitor the amount of code that is exercised when
we run our tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@
 # Ignore all logfiles and tempfiles.
 /log/*.log
 /tmp
+/coverage
 
 # Ignore structure.sql (since we've returned to schema.rb)
 # in case anyone runs db:schema:dump and checks in the result
--- a/Gemfile
+++ b/Gemfile
@@ -40,6 +40,7 @@ group :test do
   gem "launchy" # Primarily for save_and_open_page support in Capybara
   gem "rails-controller-testing"
   gem "shoulda-matchers"
+  gem "simplecov"
   gem "timecop"
   gem "webdrivers"
   gem "webmock", require: false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -136,6 +136,7 @@ GEM
     declarative (0.0.20)
     declarative-option (0.1.0)
     diff-lcs (1.3)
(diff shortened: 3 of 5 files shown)


kevindew:
It looks like this produces two different coverage reports (RSpec and Cucumer) do we not want a single one that combines them both? I've no idea if that's possible, mind.

benthorner:
It does produce two reports, but they're distinct and [automatically aggregated by SimpleCov](https://github.com/simplecov-ruby/simplecov#merging-results):

```
{
  "RSpec": {
    "coverage": {
      "/govuk/transition/app/controllers/concerns/track_mappings_progress.rb": {
        "lines": [
          1,

...

    "timestamp": 1600778396
  },
  "Cucumber Features": {
    "coverage": {
      "/govuk/transition/app/controllers/concerns/track_mappings_progress.rb": {
        "lines": [
          1,

...
```

kevindew:
Oh cool, so are we going to generate something with the file produced?

Thanks for explaining.

kevindew (approved):

benthorner:
> Oh cool, so are we going to generate something with the file produced?

I'm not sure I follow. The combined coverage is visible in the `coverage/index.html` file, although it doesn't indicate what (kind of) test is covering the code. We get the web page report by default - perhaps that was an unknown for you?

kevindew:
> > Oh cool, so are we going to generate something with the file produced?
> 
> I'm not sure I follow. The combined coverage is visible in the `coverage/index.html` file, although it doesn't indicate what (kind of) test is covering the code. We get the web page report by default - perhaps that was an unknown for you?

Sorry I wasn't very clear. I know that currently a file is created, however Jenkins doesn't do anything with it (as far as I'm aware) so I wondered if we'd be using it as part of a step or process.

benthorner:
Discussed in person and all is well: the aggregated coverage is visible in a single file; the use of SimpleCov is primarily for some minimal visibility on a local machine, although we could monitor it more comprehensively in future.

PR #1016, opened 2020-10-29
Swap `around(:all)` for `around(:example)`

lozette:
`around(:all)` is invalid syntax when placed on a context. `around(:example)`
works on contexts and has the desired effect.

Warning from rspec:
```
WARNING: `around(:context)` hooks are not supported and behave like `around(:example).
```


thomasleese (approved):
Thanks! :)

PR #1017, opened 2020-11-04
Make hosts endpoint expire immediately

thomasleese:
This should help new hosts appear in that endpoint very soon after they are imported. We're seeing an issue where the [Bouncer import job](https://github.com/alphagov/govuk-cdn-config/blob/master/lib/deploy_bouncer.rb#L109) is running too soon after the [Transition import job](https://github.com/alphagov/transition/blob/master/lib/transition/import/sites.rb) and misses some of the newly importers hosts as the controller is likely to be cached by Nginx/Varnish.

### DIFF
--- a/app/controllers/hosts_controller.rb
+++ b/app/controllers/hosts_controller.rb
@@ -1,9 +1,7 @@
-# rubocop:disable Rails/ApplicationController
-class HostsController < ActionController::Base
+class HostsController < ActionController::API
   def index
     @hosts = Host.with_cname_or_ip_address.order(:hostname)
 
     render json: HostsPresenter.new(@hosts).as_hash.to_json
   end
 end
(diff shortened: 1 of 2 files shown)


app/controllers/hosts_controller.rb
```diff
@@ -1,7 +1,7 @@
 class HostsController < ActionController::API
   def index
     @hosts = Host.with_cname_or_ip_address.order(:hostname)
-
+    expires_now
```
ChrisBAshton, line 4: Is there a risk someone could DOS the `https://transition.publishing.service.gov.uk/hosts.json` endpoint? Would it be safer to set a really tiny cache instead, e.g. 1 second?

ChrisBAshton (approved):
LGTM 👍

PR #1033, opened 2021-01-18
Remove redundant config

ChrisBAshton:
The sassLint option was removed in https://github.com/alphagov/govuk-jenkinslib/pull/80
And the rubyLintDiff option was removed in https://github.com/alphagov/govuk-jenkinslib/pull/57

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,8 +3,5 @@
 library("govuk")
 
 node {
-  govuk.buildProject(
-    sassLint: false,
-    rubyLintDiff: false
-  )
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -4,7 +4,6 @@ library("govuk")
 
 node {
   govuk.buildProject(
-    sassLint: false,
     rubyLintDiff: false
```
benthorner, line 7:
```suggestion

```

This is also now defunct.

benthorner (approved):

PR #1055, opened 2021-04-06
Revise tech docs in README to favour GOV.UK Docker

benthorner:
https://trello.com/c/Psf7EzPB/204-move-away-from-startupsh

Depends on: https://github.com/alphagov/govuk-docker/pull/476

This follows the pattern established in [1]. I've added a little
more detail about importing orgs/sites, as this isn't obvious in
the Transition Config README.

[1]: alphagov/content-publisher#2257

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,31 +3,37 @@
 Rails app for managing the transition of websites to GOV.UK. Specifically, it's for the production of and handling
 of mappings for use with [Bouncer](https://github.com/alphagov/bouncer).
 
-## Dependencies
+## Technical documentation
 
-* Redis
-* PostgreSQL 9.3+ (the app uses materialized views, which were introduced in 9.3).
-  This is included in the Trusty dev VM, which is now the default.
+This is a Ruby on Rails app, and should follow [our Rails app conventions](https://docs.publishing.service.gov.uk/manual/conventions-for-rails-applications.html).
 
-## Running the app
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -3,31 +3,31 @@
...
 
-```sh
-script/rails s
-```
-
-In development, you can run sidekiq to process background jobs:
+### Running the tests
 
-```sh
-bundle exec sidekiq -C config/sidekiq.yml
```
ChrisBAshton, line 23: Sounds like we ought to have a sidekiq worker in the [transition Docker config](https://github.com/alphagov/govuk-docker/blob/master/projects/transition/docker-compose.yml), like we did in https://github.com/alphagov/govuk-docker/pull/461. Out of scope for this PR, obviously, but I wonder if it's worth keeping this in the README as a 'gotcha' until that change to govuk-docker has been made?

README.md
```diff
@@ -3,31 +3,31 @@
...
 ```
 
-## Style guide
+### Style guide
 
 Available at /style, the guide documents how transition is using bootstrap, where the app has diverged from default
 styles and any custom styles needed to fill in the gaps.
 
-## Adding data to the transition app
+### Adding data
```
ChrisBAshton, line 25: 👍

ChrisBAshton (approved):
Approved, with one minor suggestion 👍

README.md
benthorner: Good point. The development scenarios for actually running [the worker](https://github.com/alphagov/transition/blob/63e0386498d5960a5ec26fbf1b550c533cab9fc8/app/controllers/import_batches_controller.rb#L34) seem pretty niche to me, so I'm not sure it's even worth adding it to GOV.UK Docker. I've amended to commit/message to preserve the command 👍.

PR #1090, opened 2021-08-17
Fix dependencies to allow Dependabot to raise a Rails 6.1 PR

brucebolt:
Dependabot did not raise a Rails 6.1 update PR as there were incompatible dependencies.

- Bumping `acts-as-taggable-on` to the latest version to replace the requirement for `activerecord (>= 5.0, < 6.1)` with `activerecord (>= 5.0, < 6.2)`.
- Bumping `select2-rails` to the latest patch of version 3.5 to remove the requirement for `thor (~> 0.14)`

Also adding `acts-as-taggable-on` to the dependabot configuration, so we don't get into this situation with the next Rails update.  `select2-rails` doesn't need adding as the `thor` dependency is no longer fixed to a specific version.

[Trello card](https://trello.com/c/V56Gndqj)


benjamineskola (approved):

brucebolt:
This has allowed dependabot to raise https://github.com/alphagov/transition/pull/1092.

PR #1138, opened 2022-03-04
Migrate from Jasmine gem to jasmine-browser-runner

baisa:
The Jasmine Ruby gems have been deprecated and are not compatible with Rails 7, which is now becoming a blocker.

[Trello card](https://trello.com/c/21VMx1NV/264-migrate-apps-to-use-jasmine-browser-runner)


package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "devDependencies": {
+    "jasmine-browser-runner": "^1.0.0",
+    "jasmine-core": "^4.0.1"
+  },
+  "scripts": {
+    "jasmine:prepare": "RAILS_ENV=test bundle exec rails assets:clobber assets:precompile",
+    "jasmine:ci": "yarn run jasmine:prepare && yarn run jasmine-browser-runner runSpecs",
+    "jasmine:browser": "yarn run jasmine:prepare && yarn run jasmine-browser-runner"
+  }
```
kevindew, line 10: We could use some more of the conventional package.json information in here. 

For example:

```
  "name": "transition",
  "description": "GOV.UK application to manage the transition of websites to GOV.UK",
  "private": true,
  "author": "Government Digital Service",
  "license": "MIT",
```

kevindew (commented):
Nice job - there's just a few bits to fix but it looks functionally good.

package.json
baisa: It is done 👍

kevindew (approved):
Nice one - looks good to me

PR #1240, opened 2022-11-03
Increase test coverage

brucebolt:
This increases test coverage to 97.66%, ahead of enabling this application for continuous deployment.

Screenshot from SimpleCov showing 97.66% coverage, with 2090 lines covered and 50 lines missed.

As part of this work, we identified some untested rake tasks that are no longer required, so are to be deleted.

Additionally, I have added some seed data so we're able to run the application locally without needing to replicate a database.  This made it easier to write tests for the functionality that I was less familiar with.

[Trello card](https://trello.com/c/VgRHbI98)


spec/controllers/style_controller_spec.rb
```diff
@@ -0,0 +1,18 @@
+require "rails_helper"
+
+describe StyleController do
```
BeckaL, line 3: Do we even want this style controller anymore? Had a look on the live site, seems like an odd thing to be in an app.

BeckaL (approved):
Looks good to me. Nice tidy up of some stuff too.

One question there about whether we still want the style controller, but happy to approve and let you decide on that one.

spec/controllers/style_controller_spec.rb
brucebolt: I agree, it looks useless.  I'll add it to our new Trello board for tracking app improvements that we want to consider.

PR #1314, opened 2023-05-04
Add double-lock, in preparation for simplifying lock name

KludgeKML:
Adds a second lock inside the current lock in the DistributedLock method.

This is a first step to cleaning up a little tech debt. We want to remove the environment from the lock name (not necessary because all environments have their own redis anyway, and undesirable because it requires a brakeman ignore).

However, these rake tasks are run by cron jobs and it's not necessarily possible to ensure that the job won't be running when the app is deployed, so if we just change the lock name there's a possibility two instances will try to run the task, one with the old lock, one with the new one. So here we require both locks. Once this is deployed and a day has passed we can deploy with the new lock only, safe in the knowledge that if we deploy during a rake run it'll already be locking the new lock.

https://trello.com/c/84yWQ0NJ/1662-ps-15-clean-up-brakeman-errors-following-on-from-rediscurrent-redisnew-updates

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -60,25 +60,6 @@
       "note": "This is a valid URL (see app/models/mapping.rb)"
     },
     {
-      "warning_type": "Cross-Site Request Forgery",
-      "warning_code": 7,
-      "fingerprint": "3bfc1315c7ae31a26c25849c985ceaae103923692c8e15e6875271761c2671f4",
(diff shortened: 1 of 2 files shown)


BeckaL (approved):
Thanks for the really comprehensive comment and for doing this bit of tech debt clean up ⭐  Based on what you say in the PR comment, this looks good to me.

PR #1380, opened 2023-08-31
Allow admins to delete a site and all its data

CristinaRO:
[Trello](https://trello.com/c/AQMxBocW/816-improve-the-process-for-deleting-a-site-from-transition)

This doesn't have to be the final shape of this feature, but it should be enough to enable us to think about where to go next.

Allows admins to delete a site via the interface. We may decide to use a custom permission, which Joe might already be working as part of his story, so we should coordinate this.

I've opted for a two-step, JavaScript-free process, for simplicity.

I've also opted for a GitHub-style confirmation prompt, where rather than a yes/no question, the user has to input some text (the site slug AKA abbreviation in our case) to confirm a destructive action.


features/site.feature
```diff
@@ -120,3 +120,25 @@ Scenario: Editing a site's transition date as a non-GDS Editor
...
+  And I should see the deletion confirmation message
+
+Scenario: Deleting a site as a non-Site Manager
+  Given I have logged in as a member of DCLG
+  And a site dclg exists
+  And I visit this site page
+  Then I should not see "Delete"
+  When I visit the path /sites/dclg/confirm_destroy
+  Then I should be redirected to the site dashboard
+  And I should see "Only certain users can access that."
```
JonathanHallam, line 144: Really like this test walk through, thank you :)

JonathanHallam (approved):
Awesome thank you, really like how this is looking

PR #1384, opened 2023-09-05
Add new transition form

jkempster34:
> [!NOTE]  
> This is not the finished article, this is for us to refine as we learn more about how it will be used.

https://trello.com/c/3NYAPZs4

### Add constants for global and special redirect types
We want to add a new site form. In order to do this we will need an easy way to
access these values.

We might want to consider refining these further into enums, as this would
allow us to remove some of the methods in the site model.

### Add routes for site creation
We want to add sites per organisation.

This add nested routes for site creation.

### Add site form object
We are aiming to archive Transition Config which is currently used to create
sites in Transition.

This adds a form object for the creation of a site. 

It's worth noting that the process currently moves existing hosts and aka hosts
to the new site if they already exist (see
`lib/transition/import/site_yaml_file.rb`). We aren't yet sure whether we want
to create a new UI for that functionality, or to use this form. For now, this
form does not have that functionality.

### Only validate presence of canonical host or aka if hostname is present
If hostname is not provided for the host, this validation will raise an error.

Add a condition that this validation only runs when hostname is present.

### Add site creation actions and views
We are aiming to archive Transition Config which is currently used to create
sites in Transition.

This adds the actions and views necessary for a user to add a new transition.

### Validate that host hostname is lowercase
Currently the site importer (`lib/transition/import/site_yaml_file.rb`)
downcases any hostnames.

This adds validation to the host model which will be surfaced on the new site
form.

### Only allow `Site managers` to create new sites
This adds authorization that only people with the `Site manager` role in GOV.UK
Signon can create new sites.

This is so that we can safely test new features before we officially archive
Transition Config.

### Validate that global new URL is absent if the global type is "archive"
Currently, there is validation in place to prevent the creation of sites which
have a global type of "redirect" without a global new URL, but the opposite
validation is missing.

### Improve appearance of radio buttons
In order to style these using Bootstrap, we need to provide a block to `collection_radio_buttons`.

### Nilify blanks on the Site model
The optional fields of the site form come through as empty strings when left
blank. These are then saved in the database, when ideally these would be saved
as `nil` in order to preserve the previous behaviour of adding them through
Transition Config for consistency.

We already have a module that handles the nilification of blanks, however we
need to make a couple of changes:
- We add an "except list" of attributes that we don't want to nilify, such as
  Booleans that should be false and not nil (false.blank? == true).
- We allow blank for `special_redirect_strategy` as this will now be nilified
  before save.

### Add hints to new site form inputs
This copies the guidance fro Transition Config over to the new site form.

This will need later refinement.

image

---

### DIFF
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -12,6 +12,7 @@ class Host < ApplicationRecord
   validates :hostname, hostname: true
   validates :site, presence: true
   validate :canonical_host_id_xor_aka_present, if: -> { hostname.present? }
+  validate :hostname_is_downcased, if: -> { hostname.present? }
 
   after_update :update_hits_relations, if: :saved_change_to_site_id?
 
@@ -70,5 +71,11 @@ class Host < ApplicationRecord
     host_paths.update_all(mapping_id: nil, canonical_path: nil)
     hits.update_all(mapping_id: nil)
     Transition::Import::HitsMappingsRelations.refresh!(site)
+  end
+
+  def hostname_is_downcased
+    if hostname != hostname.downcase
+      errors.add(:hostname, "must be lowercase")
+    end
   end
 end
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -11,7 +11,7 @@ class Host < ApplicationRecord
   validates :hostname, presence: true
   validates :hostname, hostname: true
   validates :site, presence: true
-  validate :canonical_host_id_xor_aka_present
+  validate :canonical_host_id_xor_aka_present, if: -> { hostname.present? }
 
   after_update :update_hits_relations, if: :saved_change_to_site_id?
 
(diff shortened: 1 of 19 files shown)


app/models/host.rb
```diff
@@ -12,6 +12,7 @@ class Host < ApplicationRecord
   validates :hostname, hostname: true
   validates :site, presence: true
   validate :canonical_host_id_xor_aka_present, if: -> { hostname.present? }
+  validate :hostname_is_downcased, if: -> { hostname.present? }
```
JonathanHallam, line 15: Can we just not downcase the host before we send it on?

JonathanHallam:
Really like this, thanks Joe, only comment is wondering whether there's a way we can let people know what all of the fields are and maybe give a link for the tna timestamp section? Maybe lets have a chat offline?

PR #1392, opened 2023-09-12
Govuk styling create form

jkempster34:


brucebolt (approved):
This all looks good to me.

Probably worth asking the #govuk-publishing-whitehall-transition team if they're also able to have a quick look, in case there's any caveats/issues they're aware of with these types of migrations.

nnagewad:
The screen shots looks great.

I know in Transition side we've been putting a required label on required items instead of optional inputs. That being said, this approach is more in line with design system guidance.

@jkempster34 be good to connect as I'm not familiar with this work...

ollietreend:
@jkempster34 it's interesting to see the [GOV.UK Form Builder](https://govuk-form-builder.netlify.app/) gem being used here. This is certainly the most Railsy way to create forms that implement GOV.UK Design System styling – but it's [not currently used in any other GOV.UK Publishing apps](https://github.com/search?q=org%3Aalphagov+govuk_design_system_formbuilder&type=code).

We did consider using this gem when beginning Whitehall's transition to the Design System – but it was decided that we should use the [govuk_publishing_components](https://components.publishing.service.gov.uk/component-guide) gem to create forms instead.

I don't think the decision was officially recorded or mandated, but it was a steer that we got from @kevindew (Lead Developer for Publishing at the time).

If I remember correctly, I think the main reasoning was that govuk_publishing_components is the established convention used across GOV.UK frontend and Publishing apps (for example, [Content Publisher](https://github.com/alphagov/content-publisher/blob/134b8df65fd88d88e6f67ebab32b5dc463baef9c/app/views/content/edit/_title_and_base_path_field.html.erb#L16-L35) and now [Whitehall](https://github.com/alphagov/whitehall/blob/c1095d2abdc3d652b8a10f06a73deb5e895e5022/app/views/admin/editions/_first_published_at.html.erb#L34-L56)). While it's not quite as Railsy, it does have the advantage of being owned in-house, so it's within our gift to adapt or change these components as needed.

It might be worth re-considering that decision if you think there's a strong argument for using the GOV.UK Form Builder gem. We'd need to consider how and when it'd be used, and have some sort of plan for how it'd be rolled out across the Publishing apps. But otherwise my recommendation would be that we [embrace the existing convention](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-161-technical-direction-in-publishing-2023.md#embrace-existing-conventions-start-new-conventions-collaboratively).

jkempster34:
> We did consider using this gem when beginning Whitehall's transition to the Design System – but it was decided that we should use the [govuk_publishing_components](https://components.publishing.service.gov.uk/component-guide) gem to create forms instead.

I didn't realise that using the form builder would be controversial! 

I think the arguments for using it will be pretty evident in this PR, I can imagine that using the components instead will make the fairly complex view quadruple in size, and I'll have to add extra logic for handling errors which then has the potential to be wrong. 

However, I definitely understand the need for consistency, so I'm sure that we will be able to convert this!

jkempster34:
> The screen shots looks great.
> 
> I know in Transition side we've been putting a required label on required items instead of optional inputs. That being said, this approach is more in line with design system guidance.
> 
> @jkempster34 be good to connect as I'm not familiar with this work...

Yep, we did discuss how best to approach that! But as you said, we followed the guidance [here](https://design-system.service.gov.uk/patterns/question-pages/#when-to-use-this-pattern).

jkempster34:
> @jkempster34 it's interesting to see the [GOV.UK Form Builder](https://govuk-form-builder.netlify.app/) gem being used here. This is certainly the most Railsy way to create forms that implement GOV.UK Design System styling – but it's [not currently used in any other GOV.UK Publishing apps](https://github.com/search?q=org%3Aalphagov+govuk_design_system_formbuilder&type=code).
> 
> We did consider using this gem when beginning Whitehall's transition to the Design System – but it was decided that we should use the [govuk_publishing_components](https://components.publishing.service.gov.uk/component-guide) gem to create forms instead.
> 
> I don't think the decision was officially recorded or mandated, but it was a steer that we got from @kevindew (Lead Developer for Publishing at the time).
> 
> If I remember correctly, I think the main reasoning was that govuk_publishing_components is the established convention used across GOV.UK frontend and Publishing apps (for example, [Content Publisher](https://github.com/alphagov/content-publisher/blob/134b8df65fd88d88e6f67ebab32b5dc463baef9c/app/views/content/edit/_title_and_base_path_field.html.erb#L16-L35) and now [Whitehall](https://github.com/alphagov/whitehall/blob/c1095d2abdc3d652b8a10f06a73deb5e895e5022/app/views/admin/editions/_first_published_at.html.erb#L34-L56)). While it's not quite as Railsy, it does have the advantage of being owned in-house, so it's within our gift to adapt or change these components as needed.
> 
> It might be worth re-considering that decision if you think there's a strong argument for using the GOV.UK Form Builder gem. We'd need to consider how and when it'd be used, and have some sort of plan for how it'd be rolled out across the Publishing apps. But otherwise my recommendation would be that we [embrace the existing convention](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-161-technical-direction-in-publishing-2023.md#embrace-existing-conventions-start-new-conventions-collaboratively).

I've now moved this to use Publishing Components. It's worth saying that it was quite a lot of effort, and there are plenty of mistakes that you can make along the way e.g., not loading values back into the inputs when invalid, showing multiple errors for a single field in the error summary (I think this is a problem in Whitehall), not linking the error messages in the summary to the fields. This is a comparison of the effort required to switch: https://github.com/alphagov/transition/commit/1d0c9cf4c0f3fc669e67b807ae46720b18305ebf

brucebolt (approved):

PR #1393, opened 2023-09-13
Add PaperTrail to sites and associated models

CristinaRO:
[Trello](https://trello.com/c/4ZnLRcFX/826-add-papertrail-to-transition-entities)

### DIFF
--- a/app/controllers/sites_controller.rb
+++ b/app/controllers/sites_controller.rb
@@ -1,6 +1,8 @@
 class SitesController < ApplicationController
   layout "admin_layout", only: %w[new create confirm_destroy destroy]
 
+  include PaperTrail::Rails::Controller
+
   before_action :find_site, only: %i[edit update show confirm_destroy destroy]
   before_action :find_organisation, only: %i[new create]
   before_action :check_user_is_gds_editor, only: %i[edit update]
--- a/app/models/hit.rb
+++ b/app/models/hit.rb
@@ -3,6 +3,8 @@ require "kaminari"
 class Hit < ApplicationRecord
   NEVER = Date.new(1970, 1, 1)
 
+  has_paper_trail
+
   belongs_to :host
   belongs_to :mapping
 
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -1,6 +1,8 @@
 require "transition/import/hits_mappings_relations"
 
 class Host < ApplicationRecord
+  has_paper_trail
+
   belongs_to :site
   has_many :hits
   has_many :host_paths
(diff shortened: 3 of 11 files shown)


app/controllers/sites_controller.rb
```diff
@@ -1,6 +1,8 @@
 require "./lib/transition/import/revert_entirely_unsafe"
 
 class SitesController < ApplicationController
+  include PaperTrail::Rails::Controller
```
jkempster34, line 4: This could possibly be added to the commit message because I had to look it up

spec/controllers/sites_controller_spec.rb
```diff
@@ -67,6 +79,37 @@ def make_request
     end
   end
 
+  describe "#update" do
+    def make_request
```
jkempster34, line 83: These might work as RSpec `subject`s, but I think thats just personal preference

app/models/host.rb
```diff
@@ -1,6 +1,8 @@
 require "transition/import/hits_mappings_relations"
 
 class Host < ApplicationRecord
+  has_paper_trail
```
jkempster34, line 4: Do these need specs the same as `Site`? Do they log the whodunnit correctly?

app/models/hit.rb
```diff
@@ -3,6 +3,8 @@
 class Hit < ApplicationRecord
   NEVER = Date.new(1970, 1, 1)
 
+  has_paper_trail
```
jkempster34, line 6: Same here, does this need specs?

features/step_definitions/site_assertion_steps.rb
```diff
@@ -98,3 +98,8 @@
 Then(/^I should be redirected to the organisation dashboard$/) do
   i_should_be_on_the_path organisation_path(@site.organisation)
 end
+
+Then(/^the last version should list "([^"]*)" as the author of the change$/) do |name|
+  expect(@site.versions.last.whodunnit).to eql(name)
```
jkempster34, line 103: I thought about doing this the other day, checking records in feature specs. I had a Google and it seem like most people would expect these to only exercise things the user can see.

jkempster34 (changes requested):
Left some comments.

The only "blocking" question I have is around the specs for hits and hosts. If they are needed, can they be put in a shared example or something?

jkempster34 (approved):
👍

features/step_definitions/site_assertion_steps.rb
jkempster34: I think you changed my mind on this. I think it's worth ignore the "ideal" standard if the test is still useful, which it was in this case!

Especially with the fact that this is not in the UI as you said

CristinaRO: Oh, just when I'd come over to your side! ;) The request spec reproduced the bug too, though, so that's all right.

CristinaRO: Hm, maybe I'll amend the commit to go into detail about that, so future devs beware of controller specs.

PR #1401, opened 2023-09-20
Validate that hostnames are unique

jkempster34:
https://trello.com/c/Aaao9833

The current import process will move a host to a new site if a site with that
same host already exists (documented in
`spec/lib/transition/import/site_yaml_file_spec.rb:177`).

We think this will not happen frequently, so have decided to disallow adding or
updating a host where the hostname is not unique.

All hostnames in the database are currently unique, so this validation can run
on create and update without invalidating any current records.

Note that this executes 2 SQL queries, whereas we could perform this in 1 if
using custom validation, however I think using the built in validation in
clearer for this use case.

image
image

---

### DIFF
--- a/app/models/host.rb
+++ b/app/models/host.rb
@@ -9,6 +9,12 @@ class Host < ApplicationRecord
   belongs_to :canonical_host, class_name: "Host"
 
   validates :hostname, presence: true
+  validates :hostname, uniqueness: {
+    message: lambda do |_object, data|
+      site = Host.find_by(hostname: data[:value]).site.abbr
+      "The hostname #{data[:value]} already exists. You must delete the #{site} site to remove it"
+    end,
+  }
   validates :hostname, hostname: true
   validates :site, presence: true
   validate :canonical_host_id_xor_aka_present, if: -> { hostname.present? }
--- a/spec/models/host_spec.rb
+++ b/spec/models/host_spec.rb
@@ -87,6 +87,23 @@ describe Host do
           expect(host.errors_on(:hostname)).to include("must be lowercase")
         end
       end
+
+      context "does not already exist" do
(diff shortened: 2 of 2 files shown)


app/models/host.rb
```diff
@@ -9,6 +9,7 @@ class Host < ApplicationRecord
   belongs_to :canonical_host, class_name: "Host"
 
   validates :hostname, presence: true
+  validates :hostname, uniqueness: { message: ->(_object, data) { "The hostname #{data[:value]} already exists" } }
```
JonathanHallam, line 12: Thanks for sorting this out. Is there any way we could let the user know what they need to do if a hostname already exists? "No" is a perfectly fine answer, just want to make sure we're making it as usable as possible

jkempster34: I think the only way to delete a host, is to delete a site (the stuff Cristina did). 

So this would either be: 

1. `The hostname www.cabinetoffice.gov.uk already exists. You must delete the cabinetoffice site to remove it` or
2. `The hostname www.cabinetoffice.gov.uk already exists. You must delete the associated site to remove it`

The problem with number 1 is that we use the default host throughout Transition to refer to a site, so it might be confusing. See pic below, the site is `cabinetoffice`, yet the heading is the default host `www.cabinetoffice.gov.uk`

image

What do you think?

jkempster34: @JonathanHallam Any thoughts on this? 🤔

JonathanHallam: Sorry, should have got back to you sooner. I think number 2. I guess bonus points if it's possible to pull a link to the associated site in?

jkempster34: We chatted about this in standup and went for option 1 so that the user can search for the existing site.

This is updated now if you can re-review please 🙏

JonathanHallam (approved):
👍

PR #1404, opened 2023-09-25
Update a site

jkempster34:
https://trello.com/c/ajngFGlz

### Edit date form

|Before|After|
|-|-|
|image|image|

### Edit site form
image

### Move site form into a partial
This form is currently used for adding a new site, but will soon be used
for editing an existing site.

Therefore we can copy the functionality.

### Add new controller for editing a sites transition date
We want to add a form to edit the details of a site. However, we already have a
form for editing a site's transition date.

This moves the action into its own controller, so that we can use the `edit`
action of the `SitesController` for the site details instead.

### Move the edit transition date form to the design system
Now that we have introduced the Design System to Transition, we want to move
over pages where possible.

Because of the complexity of the Design System's date component, we need to
back this form by a form object.

### Add form for editing a transitioned site
This was previously done by editing the information in YAML files in the
`transition-config` repo.

Adding a user interface so changes can be made without developer
intervention.

Any related hosts are unchanged if they already exist (based on hostname).  If
a host has been removed from a site, it's record is destroyed. If a hostname is
new, a new record is created.

Validation will soon be added to prevent a host being created with a hostname
that is already in use.

---


features/site.feature
```diff
@@ -130,7 +130,7 @@ Scenario: Editing a site's transition date as a non-GDS Editor
   And a site dclg exists
   And I visit this site page
   Then I should not see "Edit date"
-  When I visit the path /sites/dclg/edit
+  When I visit the path /site_dates/dclg/edit
```
brucebolt, line 146: I wonder if the path would be better as `/sites/dclg/edit_date` as that would retain the path structure associating this with a site?

It'd also mean we wouldn't need a new controller, but could instead add `edit_date` and `update_date` methods to the `SitesController`.

brucebolt (commented):
Looks good to me.  I've made one comment about whether we should keep the date edit form within the `SitesController`.

features/site.feature
jkempster34: So I've always done this and added extra actions to existing controllers. However, I read this article (and I've seen similar others) which contains good arguments for not doing that http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers/

Let me know what you think!

I do however think I should update the path at least!

jkempster34: I'm sure we can have `/sites/dclg/edit_date` or `/sites/dclg/date/edit` despite it being a different controller (if you agree that that article makes sense)

brucebolt: I can see a separate controller making make sense here.  I think `/sites/dclg/edit_date` would be my preferred URL, as there would be nothing rendered at `/sites/dclg/date`.

jkempster34: Done as a new commit 💯

brucebolt: Change looks good :+1:.  I wonder if the commit could be rebased into [`Add new controller for editing a sites transition date`](https://github.com/alphagov/transition/pull/1404/commits/0480316ad9fae66405e61c93340c278d30255ec8) so the commit history is more succinct?

jkempster34: Done! :hurtrealbad:

brucebolt (approved):

PR #1411, opened 2023-10-02
Split off orgs_sites_hosts import task

CristinaRO:
**Note**:  should be coordinated with https://github.com/alphagov/govuk-helm-charts/pull/1296

Separate the responsibilities so we can invoke separate tasks for importing organisations and importing sites and hosts, respectively.

This allows us to keep as much of the existing tests as possible, to make the path towards removing the dependency on transition-config safer and clearer.

This separation will also allow us to modify the cron job to only invoke the importing of organisations from whitehall.

[Trello](https://trello.com/c/toGB61iI/809-remove-cron-job-that-imports-site-data-from-transition-config-to-transition-then-archive-the-transition-config-repo)

### DIFF
--- a/lib/tasks/import/orgs_sites_hosts.rake
+++ b/lib/tasks/import/orgs_sites_hosts.rake
@@ -1,9 +1,14 @@
 require "transition/import/orgs_sites_hosts"
 
 namespace :import do
-  desc "Import all Organisations, and Sites and Hosts from the given filename or mask"
-  task :orgs_sites_hosts, [:filename_or_mask] => :environment do |_, args|
-    Transition::Import::OrgsSitesHosts.from_yaml!(args.filename_or_mask)
+  desc "Import all Organisations"
+  task organisations: :environment do
+    Transition::Import::Organisations.from_yaml!
+  end
+
+  desc "Import all Sites and Hosts from the given filename or mask"
+  task :sites_hosts, [:filename_or_mask] => :environment do |_, args|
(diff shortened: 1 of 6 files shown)


lib/tasks/import/orgs_sites_hosts.rake
```diff
@@ -1,9 +1,14 @@
...
 namespace :import do
-  desc "Import all Organisations, and Sites and Hosts from the given filename or mask"
-  task :orgs_sites_hosts, [:filename_or_mask] => :environment do |_, args|
-    Transition::Import::OrgsSitesHosts.from_yaml!(args.filename_or_mask)
+  desc "Import all Organisations"
+  task organisations: :environment do
+    Transition::Import::Organisations.from_yaml!
+  end
+
+  desc "Import all Sites and Hosts from the given filename or mask"
```
brucebolt, line 9: I _think_ we'll be able to remove this rake task (and the associated code) too, since we will no longer need to import anything from Transition Config.

But this looks totally fine if you're planning on the code removal being done in two stages.

brucebolt (approved):

lib/tasks/import/orgs_sites_hosts.rake
CristinaRO: I'd rather do it in two stages, since it looks like we'll be able to get rid of `.notmodules` and related things as well 🤔

brucebolt: Seems like a good approach 👍

PR #1416, opened 2023-10-05
Remove transition-config leftovers

CristinaRO:
https://trello.com/c/4pPxycMw/810-identify-all-of-the-things-related-to-transition-config-in-transition-and-remove-them

Since Transition no longer imports data from transition-config, remove the remaining code that was used for that.

I think the commits are fairly self-contained and I've tried to describe what's changing and why it's changing as I've gone along, so I won't repeat that here. Let me know if it's not as clear as I think it is.


lib/transition/import/whitehall_orgs.rb
```diff
@@ -8,13 +8,6 @@ module Import
...
       include Enumerable
 
-      # org_yaml_path is only for test usage, so that we can pass in a fixture
-      # rather than calling a live API. In production this always calls the
-      # live API.
-      def initialize(org_yaml_path = nil)
-        @org_yaml_path = org_yaml_path
-      end
-
       def organisations
```
jkempster34, line 11: I wonder if `all` might have been a better method name here (again predates this PR sorry)

PR #1439, opened 2023-10-30
Update rails 7.1.1

jkempster34:
https://trello.com/c/ZY91ifiL

### Run Rails update task
As part of updating Rails to 7.1, we should run the Rails update task.

This runs the task, updating the config where necessary.

### Pass class as a keyword argument
Passing the class to serialize as a positional argument is deprecated in Rails
7.1.

### Set serializer coder to YAML
Prior to Rails 7.1, this defaulted to YAML. Now we must specify it.

### Load secrets into credentials hash to maintain existing behaviour
This copies the approach taken by the Account API team
(https://github.com/alphagov/account-api/pull/732).

### Move class method into CanonicalFilter class
Currently, the `canonical_filter` class method is located in the
`View::Mappings` module. This is called in the `Filter` class (which is also
located in `View::Mappings`) because the entire
`view/mappings/canonical_filter` file is loaded manually with `require`.

However, in order to use the Rails 7.1defaults, this require must be removed as
we can't manually require files that are managed by the autoloader. This means
that `canonical_filter` will no longer be available in the `Filter` class.

This moves the functionality of `canonical_filter` into a factory method on the
`CanonicalFilter` class.

### Update class names with abbreviations
Using the Rails 7.1 defaults, we see an errors where file names are not what is
expected by the autoloader.

All of the examples here are due to the capitalisation of abbreviations.

This aims for some consistency (all acronyms are capitalised) by either
updating the Zetwerk config or by updating the class name, depending on whether
Zeitwork is natively aware of the abbreviation.

### Remove manual requires for files managed by the autoloader
In Rails 7.1, the `add_autoload_paths_to_load_path` config is defaulted to
false, which means we can't manually require files that are managed by the
autoloader.

### Use Rails 7.1 defaults

---


app/models/user.rb
```diff
@@ -5,7 +5,7 @@ class User < ApplicationRecord
   has_many :bulk_add_batches
   has_many :import_batches
 
-  serialize :permissions, type: Array
+  serialize :permissions, type: Array, coder: YAML
```
JonathanHallam, line 8: Does this functionally change anything? I'm not familiar with coder, I'm basically worried that if we're trying to pass anything but yaml at the moment then this might fail? But this comment may be really stupid because I don't know what I'm talking about!

JonathanHallam (approved):
Tiny comment that I assume is non blocking. Thanks for this :)

PR #1468, opened 2023-12-07
Accessibile autocomplete multiselect

jkempster34:
### Load the selected extra organisations into the site form correctly
The values of the extra organisations select component correspond to the record
IDs.

### Install Yarn as part of Dockerfile
This is required now that we want to manage our dependency on the
accessible-autocomplete-multiselect component with Yarn.

### Add node modules to assets path
Because we are using Yarn to manage our dependency on the
accessible-autocomplete-multiselect component, Rails needs to know to look in
node modules for assets.

### Use autocomplete multiselect component for extra organisations
Currently, extra organisations on the site form use a multi-select element.

This is quite hard to use for long lists of organisations, so we want to use
the autocomplete multiselect component instead.

With JavaScript disabled, this will fall back to the multi-select element.

---

This application is owned by the publishing platform team. Please let us know in #govuk-publishing-platform when you raise any PRs.

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.

PR #1471, opened 2023-12-12
Accessible autocomplete

jkempster34:
https://trello.com/c/iSpvh7PQ

### Load the selected extra organisations into the site form correctly
The values of the extra organisations select component correspond to the record
IDs.

### Use autocomplete multiselect component for extra organisations
Currently, extra organisations on the site form use a multi-select element.

This is quite hard to use for long lists of organisations, so we want to use
the autocomplete multiselect component instead.

With JavaScript disabled, this will fall back to the multi-select element.

|Before (or with JavaScript disabled)|After|
|-|-|
|image|image|

---


app/assets/javascripts/accessible-autocomplete.min.js
```diff
@@ -0,0 +1,2 @@
...createElement)("div",{style:{border:"0",clip:"rect(0 0 0 0)",height:"1px",marginBottom:"-1px",marginRight:"-1px",overflow:"hidden",padding:"0",position:"absolute",whiteSpace:"nowrap",width:"1px"}},(0,b.createElement)("div",{id:t+"__status--A",role:"status","aria-atomic":"true","aria-live":"polite"},!h&&d&&f?_:""),(0,b.createElement)("div",{id:t+"__status--B",role:"status","aria-atomic":"true","aria-live":"polite"},h||!d||f?"":_))},e}(b.Component);(t["default"]=r).defaultProps={tQueryTooShort:function(e){return"Type in "+e+" or more characters for results"},tNoResults:function(){return"No search results"},tSelectedOption:function(e,t,n){return e+" "+(n+1)+" of "+t+" is highlighted"},tResults:function(e,t){return e+" "+(1===e?"result":"results")+" "+(1===e?"is":"are")+" available. "+t}}},function(e,t,n){"use strict";t.__esModule=!0,t["default"]=void 0;var o=n(5),r=function i(e){var t=e.className;return(0,o.createElement)("svg",{version:"1.1",xmlns:"http://www.w3.org/2000/svg",className:t,focusable:"false"},(0,o.createElement)("g",{stroke:"none",fill:"none","fill-rule":"evenodd"},(0,o.createElement)("polygon",{fill:"#000000",points:"0 0 22 0 11 17"})))};t["default"]=r}])["default"]});
```
brucebolt, line 1: Instead of including the minified JavaScript and CSS (which is mostly unreadable), would we be able to include the component in the same way it has been done in Whitehall?  Then we won't need to try and keep these up-to-date ourselves and can use dependabot to that for us.

E.g. https://github.com/alphagov/whitehall/blob/3fdc3561228862058b831527c3bab22ae24a73ff/yarn.lock#L180-L184

jkempster34: I should have said in the PR description, I tried this https://github.com/alphagov/transition/pull/1468

However, I'm struggling to install packages from repos with Yarn v3 (Transition uses Yarn 3, Whitehall uses Yarn 1), so I opted for the easiest solution which was this https://github.com/alphagov/accessible-autocomplete?tab=readme-ov-file#as-a-plain-javascript-module

You can see the error by running `yarn set version 3.4.1 && yarn add accessible-autocomplete-multiselect@alphagov/accessible-autocomplete-multiselect` vs `yarn set version 1.22.19 && yarn add accessible-autocomplete-multiselect@alphagov/accessible-autocomplete-multiselect`

Maybe I'll raise this in dev chat!

jkempster34:
Closing because the Publishing Experience team have made the choice to stop maintaining Accessible Autocomplete Multiselect and we might want to consider a different pattern such as the "Add another" pattern

See https://trello.com/c/iSpvh7PQ

PR #1473, opened 2023-12-13
Change URLs to use IDs instead of abbreviations

brucebolt:
Users are currently asked for an abbreviation for sites when they set them up.  This serves no purpose, other than creating a slug for use in URLs within the Transition tool.  It is not used anywhere externally.

This is a waste of user's time and may cause confusion as to what actually needs to be entered as an abbreviation.

Therefore moving away from abbreviations and switching to IDs within URLs, rather than the abbreviation slug.  The field is removed from the create/edit form, however we maintain access to URLs that already contained a slug.

[Trello card](https://trello.com/c/HaVI5NDq)

### DIFF
--- a/app/controllers/site_dates_controller.rb
+++ b/app/controllers/site_dates_controller.rb
@@ -25,7 +25,7 @@ class SiteDatesController < ApplicationController
 private
 
   def find_site
-    @site = Site.find_by!(abbr: params[:site_id])
+    @site = Site.find_by_abbr_or_id(params[:site_id])
   end
 
   def update_params
--- a/app/controllers/sites_controller.rb
+++ b/app/controllers/sites_controller.rb
@@ -58,7 +58,7 @@ class SitesController < ApplicationController
 private
 
   def find_site
-    @site = Site.find_by!(abbr: params[:id])
+    @site = Site.find_by_abbr_or_id(params[:id])
   end
 
   def find_organisation
--- a/app/controllers/sites_controller.rb
+++ b/app/controllers/sites_controller.rb
@@ -46,7 +46,7 @@ class SitesController < ApplicationController
   end
 
   def destroy
-    @delete_site_form = DeleteSiteForm.new(abbr: @site.abbr, **destroy_params)
+    @delete_site_form = DeleteSiteForm.new(id: @site.id, **destroy_params)
 
     if @delete_site_form.save
       redirect_to organisation_path(@site.organisation), flash: { success: "The site and all its data have been successfully deleted" }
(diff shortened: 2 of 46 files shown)


spec/controllers/sites_controller_spec.rb
```diff
@@ -66,17 +66,22 @@ def make_request
...
-      it "displays the form" do
+      it "displays the form when requesting using an abbreviation" do
         get :edit, params: { id: site.abbr }
         expect(response.status).to eql(200)
       end
+
+      it "displays the form when requesting using an ID" do
+        get :edit, params: { id: site.id }
+        expect(response.status).to eql(200)
+      end
```
mike29736, line 77: Since we're offering backwards compatibility, shouldn't we also add an example like this to the specs for all the other controllers we're changing?

spec/helpers/hits_helper_spec.rb
```diff
@@ -92,25 +92,25 @@
     end
 
     context "when a site is present" do
-      let(:site) { build :site, abbr: "site_abbr" }
+      let(:site) { create :site, abbr: "site_abbr" }
```
mike29736, line 95: Can the `abbr` be removed now?

mike29736 (approved):
A few small comments, no deal breakers

app/controllers/sites_controller.rb
```diff
@@ -58,7 +58,7 @@ def destroy
 private
 
   def find_site
-    @site = Site.find_by!(abbr: params[:id])
+    @site = Site.find_by(abbr: params[:id]) || Site.find_by!(id: params[:id])
```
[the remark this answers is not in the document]
brucebolt: Ooh, yes, will change this.

app/controllers/site_dates_controller.rb
```diff
@@ -25,7 +25,7 @@ def update
 private
 
   def find_site
-    @site = Site.find_by!(abbr: params[:site_id])
+    @site = Site.find_by(abbr: params[:site_id]) || Site.find_by!(id: params[:site_id])
```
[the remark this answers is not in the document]
brucebolt: Nice idea.  Will add a class method.

spec/controllers/sites_controller_spec.rb
brucebolt: Yes, will add a test to each as it looks like I forgot to do that.

spec/helpers/hits_helper_spec.rb
brucebolt: Yes, that's a good idea, so I've done this in all cases where we're not using the `abbr` in a test.

It also revealed some other code that depended on `abbr` but wasn't caught by other tests, so I've updated those to use the ID too.

PR #1618, opened 2024-06-13
Sanitize and cleanup invalid UTF-8 characters

brucebolt:
We have been seeing exceptions `string contains null byte (ArgumentError)` logged in Sentry for this application.

These are a result of an automated scan being run by Cyber Security which includes a null byte in the request (we believe it is in the session cookie).  No vulnerability is present, but the resulting error thrown by the application has resulted in a lot of noise in our Sentry alerts.

Adding a test to reproduce this problem so we are able to fix it.  Sanitising request URIs, headers and cookies to prevent the exception being raised again.

[Trello card](https://trello.com/c/B0hXNjfb)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,6 +22,7 @@ gem "optic14n" # Ideally version should be synced with bouncer
 gem "paper_trail"
 gem "pg"
 gem "plek"
+gem "rack-utf8_sanitizer"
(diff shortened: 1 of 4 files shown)


Gweaton (approved):
Nice one

PR #1710, opened 2024-10-03
Update label for aliases field

brucebolt:
I was getting a cryptic error of ` is an invalid domain` when trying to add an alias domain to an existing record.

Updating the label to make it clear you cannot include spaces between the domains, else we try adding the space as a domain.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -86,7 +86,7 @@ en:
         special_redirect_strategy_options:
           via_aka: The supplier is redirecting some paths to our aka domain
           supplier: The supplier is managing redirects to gov.uk. No traffic comes through Bouncer for this site.
-        aliases: This is a list of alias domains. Enter as a comma-separated list.
+        aliases: This is a list of alias domains. Enter as a comma-separated list, without spaces.
   site:
     confirm_destroy:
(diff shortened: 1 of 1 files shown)


yndajas (approved):
Is it worth wrapping the domain string in quotes in the error message as well?

How did it end up with an empty string or a single space character though if it's just splitting on commas? I would've thought e.g. `"a.com", "b.com"` would lead to `"a.com"` and `" b.com"` - was your input in some other form?

I suppose we could handle this with a `strip`, but it looks like there's a lot of splitting in this repo without whitespace handling, so maybe not for now

brucebolt:
That's a good idea, but I was trying to avoid making any code changes, as we've got a card somewhere on the preparing board to make changes to this form/process so will save any effort for whenever we do that.

> How did it end up with an empty string or a single space character though if it's just splitting on commas? I would've thought e.g. `"a.com", "b.com"` would lead to `"a.com"` and `" b.com"` - was your input in some other form?

Maybe the error handling is somewhat broken?  I would've expected ` b.com` to be included in the error too.

We can revisit all of this whenever we update the form (it's rarely used right now, so probably not worth investing too much effort).

PR #1714, opened 2024-10-08
Configure hosts for application

brucebolt:
Note: the healthcheck endpoints are requested by IP, not domain, so we need to specifically exclude them from the protection.<br><br>[Trello card](https://trello.com/c/frqFN7D8)


yndajas (approved):
Might be good to explain the reason for this change in brief in the commit message

PR #1725, opened 2024-10-17
Remove breadcrumbs to appease WCAG

callumknights:
Trello: https://trello.com/c/vB0leYNw/1393-fix-breadcrumbs-wcag-fails-in-transition

After: 
image

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -56,14 +56,6 @@
     %>
   <% end %>
 
-  <% breadcrumbs.tap do |links| %>
-    <% if links.any? %>
-      <ul class='breadcrumb'>
-        <% links.each do |link| %><%= crumb_li(link.text, link.url, link.current?) %><% end %>
-      </ul>
-    <% end %>
-  <% end %>
-
   <%= yield %>
   <%= render partial: 'shared/go_to' %>
 <% end %>


app/views/layouts/admin_layout.html.erb
```diff
@@ -11,18 +11,6 @@
...
-    <% if breadcrumbs.links.any? %>
-      <%= render "govuk_publishing_components/components/breadcrumbs", {
-        collapse_on_mobile: true,
-        breadcrumbs:
-          breadcrumbs
-            .links
-            .reject { |link| link.current? }
-            .map { |link| { title: link.text, url: link.url } }
-      } %>
-    <% end %>
```
yndajas, line 24: **issue:** scope

Transition is using two different layouts depending on the page: one is on the Design System (mostly for newer pages), while the other is using the Bootstrap gem (mostly for older, often more complex pages)

This file is the layout that uses the Design System, and the breadcrumbs are outside the `main` element here (the `main` element is just below), so I don't think we need to remove the breadcrumbs here. The other file `application.html.erb` is the one using the Bootstrap layout, so that's the only one from which we should need to remove breadcrumbs

callumknights: Cool, I'll correct that

yndajas (approved):

PR #1811, opened 2025-01-07
Update Rails default config for 8.0

mike3985:
I forgot to do this the other day [when I upgraded the Rails gems](https://github.com/alphagov/transition/pull/1797).

https://trello.com/c/wmR6DRXB/1549-properly-complete-rails-8-upgrades-for-transition


public/406-unsupported-browser.html
JonathanHallam, line 1: I know it comes from the update so I'm fine with it, but this change seems way less clean

bin/setup
JonathanHallam, line 1: You mention in this commit that it'll reduce noise when you run the update, but have you actually run the update, because I don't see an update commit and running it locally gets me a load of changes that aren't included here

yndajas (approved):

PR #2291, opened 2026-09-01
[PP-7949] Bump govuk_publishing_components to v68.3.0

callumknights:
I've tried to keep this as lightweight as possible, as despite the component
guide stating many more Js components than required here are needed, none of those
changed with this release, and the app did fine without them before.
So this PR merely:
- Bumps the govuk_publishing_components version
- Removes the `layout-header` es6 component, as it's no longer needed.

This application is owned by the Content APIs team. Please let us know in [#govuk-content-apis](https://gds.slack.com/archives/C03D792LYJG) when you raise any PRs.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -289,7 +289,7 @@ GEM
     govuk_personalisation (1.2.0)
       plek (>= 1.9.0)
       rails (>= 6, < 9)
-    govuk_publishing_components (67.0.1)
+    govuk_publishing_components (68.3.0)
       chartkick
       govuk_app_config
       govuk_personalisation (>= 0.7.0)
--- a/app/assets/javascripts/es6-components.js
+++ b/app/assets/javascripts/es6-components.js
@@ -10,6 +10,5 @@
 //= require govuk_publishing_components/components/button
 //= require govuk_publishing_components/components/checkboxes
 //= require govuk_publishing_components/components/error-summary
-//= require govuk_publishing_components/components/layout-header
 //= require govuk_publishing_components/components/radio
 //= require govuk_publishing_components/components/skip-link


config/initializers/dartsass.rb
```diff
@@ -7,3 +7,9 @@
 Rails.application.config.dartsass.builds = all_stylesheets
 
 Rails.application.config.dartsass.build_options << " --quiet-deps"
+
+if Rails.env.development?
+  Rails.application.config.dartsass.builds.merge!(
+    GovukPublishingComponents::Config.component_guide_stylesheet,
+  )
+end
```
yndajas: **suggestion:** I know this was in the Signon upgrade, but it might be good to explain what this is for. Does it allow us to see the component guide at `/component-guide` in dev?

Gemfile.lock
```diff
@@ -289,7 +289,7 @@ GEM
     govuk_personalisation (1.2.0)
       plek (>= 1.9.0)
       rails (>= 6, < 9)
-    govuk_publishing_components (67.0.1)
+    govuk_publishing_components (68.3.0)
```
yndajas, line 292: **question:** did you run `bundle update govuk_publishing_components`? Some of the transitive dependencies were updated for me 👀

config/initializers/dartsass.rb
yndajas: Didn't work in Signon. Maybe there's an extra step to mount it

Gemfile.lock
callumknights: I just added the `--conservative` flag

callumknights: I tend to do this habitually for most upgrades, and seeing if the resulting upgrade is stable without doing the transitive dependencies first.

config/initializers/dartsass.rb
callumknights: This comment made me go back and double check, and it works fine without, so I think it's unnecessary. And no notes in the change logs about it either.

Gemfile.lock
yndajas: Ah interesting. That sounds pretty sensible. I'll do the same!

yndajas (approved):

PR #2292, opened 2026-09-02
[PP-7949] Switch on `use_es6_components` config option

callumknights:
Correctly enables the `use_es6_components` config option, which is needed as app uses the `layout_for_admin` template.

Follow up to #2291 
https://github.com/alphagov/govuk_publishing_components/blob/c451fbe9b0ab94b7cf4bb6275019a53a8b51beac/docs/install-and-use.md#new-use_es6_components-config-option-optional

This application is owned by the Content APIs team. Please let us know in [#govuk-content-apis](https://gds.slack.com/archives/C03D792LYJG) when you raise any PRs.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -1,7 +1,6 @@
 <% content_for :head do %>
   <%= stylesheet_link_tag "legacy_layout", :media => "all" %>
   <%= javascript_include_tag "legacy_layout" %>
-  <%= javascript_include_tag "es6-components", type: "module" %>
   <%= csrf_meta_tag %>
   <%= render "layouts/google_tag_manager" %>
   <%= yield :extra_headers %>
--- a/config/initializers/govuk_publishing_components.rb
+++ b/config/initializers/govuk_publishing_components.rb
@@ -1,0 +1,3 @@
+GovukPublishingComponents.configure do |config|
+  config.use_es6_components = true
+end


geek4good (approved):

yndajas (commented):
It might be useful to include some of the detail from the Support commit message here: https://github.com/alphagov/support/pull/2101/changes/40aa2f6f852a638e86e87e74115c72c22e58806f

In short:
- in https://github.com/alphagov/transition/commit/7820bee3dbc776cbe2781af380c33923d55407a5, the ES6 components were included in the legacy Bootstrap layout instead of the Design System one (`app/views/layouts/admin_layout.html.erb`) - the latter is where they would actually have an effect
- we could move the include tag to the correct layout, but instead we're deferring to the gem to decide how to include it by setting the config option

callumknights:
Thanks @yndajas. I've included a short summary of that in the commit message.

brucebolt:
Just as a pedantic point, could you change the commit message title to include the action (e.g. `Correctly source es6-components.js`, as per @yndajas's commit message), as that'll be more useful in future rather than trying to understand what PR #2291 was related to.

yndajas:
Couple of other pedantic points, sorry!

- "Correctly enables the `use_es6_components` config option, which is needed as app uses the `layout_for_admin` template." - it's not technically _needed_, since you could just move the include tag into the correct layout, per https://github.com/alphagov/govuk_publishing_components/blob/main/docs/install-and-use.md#create-a-separate-js-bundle-for-govuk-frontend-components-optional. The config option is just an alternative (and I think preferable) approach
- "Instead of moving the layout" -> "Instead of moving _the include tag_ to the correct layout"

callumknights:
Pedantic points are good! I'd rather make the edits and make sure everything is clear for future readers, so thanks both!

I've addressed both points so hopefully this is all good to go :)

